Issue #13, opened 2023-10-28
'AfWindows' parameter occurs error.


Woojin-Crive:
## Error
```
terminate called after throwing an instance of 'std::length_error'
  what():  cannot create std::vector larger than max_size()
```

## Debug
```
"error " << id->type() << ", " << id->name() << ", " << info.def().toString() << ", " << info.min().toString() << ", " << info.max().toString());
error 7, AfWindows, (0, 0)/0x0, (0, 0)/0x0, (65535, 65535)/65535x65535
```

Error is caused when the below code is executed.
```
const rclcpp::ParameterValue value =
      cv_to_pv(clamp(info.def(), info.min(), info.max()), extent);
```

christianrauch:
Can you please provide the details (libcamera version, camera hardware, etc.) and steps to reproduce this?

Woojin-Crive:
i am using
libcamera v0.1.0+103-02eea04
raspberry camera module v3 (imx708)
latest raspberry pi os (Debian GNU/Linux 12 (bookworm))

just running the given code can reproduce the error.

christianrauch:
Can you provide a proper gdb backtrace to see where exactly the exception is thrown and what the content of the variables is?

Can you show the content of `extent` and also show what [`std::vector<T,Allocator>::max_size`](https://en.cppreference.com/w/cpp/container/vector/max_size) is? I assume the exception comes from https://github.com/christianrauch/camera_ros/blob/adfce05acb769a50f73fd9afed0c534e2927995b/src/cv_to_pv.cpp#L91 because `extent` is larger than `std::vector<T>::max_size` or negative.

Woojin-Crive:
Sorry for the dirty pull request and commits. I am not familiar with using git..
I closed the pull request.

If you give me some time. I will try to get the backtrace information.

Woojin-Crive:
extent => 18446744073709551615
max size of std::vector<T> => 576460752303423487

As you expected, `extent` is larger than `std::vector<T>::max_size `.

christianrauch:
I see the issue now. `AfWindows` is defined as `Control<Span<const Rectangle>>`, i.e. it takes a dynamic list of "windows", and happens to be the only control that has no fixed extent (length) specified. The default value for the `extent` is `libcamera::dynamic_extent`which is simply the maximum possible value the type `std::size_t` can take.

Can you check if https://github.com/christianrauch/camera_ros/pull/15 fixes the issue for you?

[...]

PR #14, opened 2023-10-28
AfWindows

Woojin-Crive:
#13

Avoid AfWindows parameter to work properly.
Raspberry pi 4 b with latest os and rapicam v3

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,6 +60,12 @@ target_link_libraries(camera_component ${libcamera_LINK_LIBRARIES} utils)
 install(TARGETS camera_component
   DESTINATION lib)
 
+# install the launch directory
+install(DIRECTORY
+  launch
+  DESTINATION share/${PROJECT_NAME}/
+)
+
 if(BUILD_TESTING)
   set(ament_cmake_clang_format_CONFIG_FILE "${CMAKE_SOURCE_DIR}/.clang-format")
   find_package(ament_lint_auto REQUIRED)
--- a/launch/camera_launch.py
+++ b/launch/camera_launch.py
@@ -7,8 +7,8 @@ from launch_ros.actions import Node
 def generate_launch_description():
     return LaunchDescription(
         [
-            DeclareLaunchArgument("width", default_value="1280", description="Camera image width"),
-            DeclareLaunchArgument("height", default_value="720", description="Camera image height"),
+            DeclareLaunchArgument("width", default_value="640", description="Camera image width"),
+            DeclareLaunchArgument("height", default_value="320", description="Camera image height"),
             DeclareLaunchArgument("log_level", default_value="debug", description="Log level (e.g., debug)"),
             # You can add other launch arguments as needed.
             # Run the camera_node with the specified parameters.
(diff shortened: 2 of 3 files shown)


src/CameraNode.cpp
```diff
@@ -164,11 +164,20 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options) : Node("camera", opti
...
+  rclcpp::PublisherOptions pub_options;
+  pub_options.qos_overriding_options = qos_override_options;
   // publisher for raw and compressed image
-  pub_image = this->create_publisher<sensor_msgs::msg::Image>("~/image_raw", 1);
+  pub_image = this->create_publisher<sensor_msgs::msg::Image>("~/image_raw", 1, pub_options);
   pub_image_compressed =
-    this->create_publisher<sensor_msgs::msg::CompressedImage>("~/image_raw/compressed", 1);
-  pub_ci = this->create_publisher<sensor_msgs::msg::CameraInfo>("~/camera_info", 1);
+    this->create_publisher<sensor_msgs::msg::CompressedImage>("~/image_raw/compressed", 1, pub_options);
+  pub_ci = this->create_publisher<sensor_msgs::msg::CameraInfo>("~/camera_info", 1, pub_options);
```
christianrauch, line 180: Why are you changing the default QoS settings? If you need to change the QoS settings, you can also do this via parameters. See `qos_overrides` in the full list of parameters: `ros2 param list`.

CMakeLists.txt
```diff
@@ -60,6 +60,12 @@ target_link_libraries(camera_component ${libcamera_LINK_LIBRARIES} utils)
 install(TARGETS camera_component
   DESTINATION lib)
 
+# install the launch directory
+install(DIRECTORY
+  launch
+  DESTINATION share/${PROJECT_NAME}/
+)
+
```
christianrauch, line 68: I would prefer not to add a launch file in this repo since the node already uses sensible defaults provided by libcamera.

launch/camera_launch.py
```diff
@@ -0,0 +1,32 @@
...
+from launch.actions import DeclareLaunchArgument, LogInfo
+from launch.substitutions import LaunchConfiguration
+from launch_ros.actions import Node
+
+
+def generate_launch_description():
+    return LaunchDescription(
+        [
+            DeclareLaunchArgument("width", default_value="1280", description="Camera image width"),
+            DeclareLaunchArgument("height", default_value="720", description="Camera image height"),
```
christianrauch, line 11: These defaults may not work for every camera.

launch/camera_launch.py
```diff
@@ -0,0 +1,32 @@
...
+                    {"width": LaunchConfiguration("width")},
+                    {"height": LaunchConfiguration("height")},
+                    {"qos_overrides./camera/image_raw.publisher.reliability": "best_effort"},
+                    {"qos_overrides./camera/image_raw/compressed.publisher.reliability": "best_effort"},
+                    {"qos_overrides./camera/camera_info.publisher.reliability": "best_effort"},
+                ],
+                remappings=[],  # Add remappings if necessary.
+                arguments=["--log-level", LaunchConfiguration("log_level")],
+            ),
+            # You can add additional nodes or actions here as needed.
```
christianrauch, line 30: Most of these comments are redundant as they apply to generic launch file configurations, which are already documented in the official documentation.

christianrauch (changes requested):
As far as I can tell, this PR does not fix the original problem. It is just ignoring a setting. Additionally, it is not well documented in the commit message what is actually happening and why these changes are required.

Woojin-Crive:
Oh sorry for the dirty code.
I was just testing for private use.

christianrauch:
You seem to add unrelated commits to this PR. Are you intending to address your issue with the `AfWindows` control in this PR?

PR #23, opened 2024-02-21
functionality for compression quality

emrekuru97:
Implemented paramtric compression quality only if strem from camera is raw, it can be done when the strem from camera is compressed too but  i tried to avoid double compression.

### DIFF
--- a/src/CameraNode.cpp
+++ b/src/CameraNode.cpp
@@ -33,6 +33,7 @@
 #include <map>
 #include <memory>
 #include <mutex>
+#include <opencv2/imgcodecs.hpp>
 #include <optional>
 #include <rcl/context.h>
 #include <rcl_interfaces/msg/detail/floating_point_range__struct.hpp>
@@ -48,6 +49,7 @@
 #include <rclcpp/qos_event.hpp>
 #include <rclcpp/time.hpp>
 #include <rclcpp_components/register_node_macro.hpp>
+#include <sensor_msgs/image_encodings.hpp>
 #include <sensor_msgs/msg/detail/camera_info__struct.hpp>
 #include <sensor_msgs/msg/detail/compressed_image__struct.hpp>
 #include <sensor_msgs/msg/detail/image__struct.hpp>
@@ -60,6 +62,7 @@
 #include <unordered_map>
 #include <utility>
 #include <vector>
+namespace enc = sensor_msgs::image_encodings;
 namespace rclcpp
 {
 class NodeOptions;
@@ -108,6 +111,8 @@ private:
   // keep track of set parameters
   ParameterMap parameters_full;
   std::mutex parameters_lock;
+  // compression quality parameter
+  std::atomic_uint8_t jpeg_quality;
 
   void
   declareParameters();
@@ -164,6 +169,18 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options) : Node("camera", opti
   // camera ID
   declare_parameter("camera", rclcpp::ParameterValue {}, param_descr_ro.set__dynamic_typing(true));
 
+  rcl_interfaces::msg::ParameterDescriptor jpeg_quality_description;
+  jpeg_quality_description.name = "jpeg_quality";
+  jpeg_quality_description.type = rcl_interfaces::msg::ParameterType::PARAMETER_INTEGER;
+  jpeg_quality_description.description = "Image quality for JPEG format";
+  jpeg_quality_description.read_only = false;
+  rcl_interfaces::msg::IntegerRange jpeg_range;
+  jpeg_range.from_value = 1;
+  jpeg_range.to_value = 100;
+  jpeg_range.step = 1;
+  jpeg_quality_description.integer_range = {jpeg_range};
+  // default to 95
+  jpeg_quality = declare_parameter<uint8_t>("jpeg_quality", 95, jpeg_quality_description);
   // publisher for raw and compressed image
   pub_image = this->create_publisher<sensor_msgs::msg::Image>("~/image_raw", 1);
   pub_image_compressed =
@@ -481,6 +498,40 @@ CameraNode::declareParameters()
   set_parameters(parameters_init_list);
 }
 
+// The following function "compressImageMsg" is adapted from "CvImage::toCompressedImageMsg"
+// (https://github.com/ros-perception/vision_opencv/blob/066793a23e5d06d76c78ca3d69824a501c3554fd/cv_bridge/src/cv_bridge.cpp#L512-L535)
+// and covered by the BSD-3-Clause licence.
+void
+compressImageMsg(const sensor_msgs::msg::Image &source,
+                 sensor_msgs::msg::CompressedImage &destination,
+                 const std::vector<int> &params = std::vector<int>())
+{
+  std::shared_ptr<CameraNode> tracked_object;
+  cv_bridge::CvImageConstPtr cv_ptr = cv_bridge::toCvShare(source, tracked_object);
+
+  destination.header = source.header;
+  cv::Mat image;
+  if (cv_ptr->encoding == enc::BGR8 || cv_ptr->encoding == enc::BGRA8 ||
+      cv_ptr->encoding == enc::MONO8 || cv_ptr->encoding == enc::MONO16)
+  {
+    image = cv_ptr->image;
+  }
+  else {
+    cv_bridge::CvImagePtr tempThis = std::make_shared<cv_bridge::CvImage>(*cv_ptr);
+    cv_bridge::CvImagePtr temp;
+    if (enc::hasAlpha(cv_ptr->encoding)) {
+      temp = cv_bridge::cvtColor(tempThis, enc::BGRA8);
+    }
+    else {
+      temp = cv_bridge::cvtColor(tempThis, enc::BGR8);
+    }
+    image = temp->image;
+  }
+
+  destination.format = "jpg";
+  cv::imencode(".jpg", image, destination.data, params);
+}
+
 void
 CameraNode::requestComplete(libcamera::Request *request)
 {
@@ -523,7 +574,7 @@ CameraNode::requestComplete(libcamera::Request *request)
 
       // compress to jpeg
       if (pub_image_compressed->get_subscription_count())
-        cv_bridge::toCvCopy(*msg_img)->toCompressedImageMsg(*msg_img_compressed);
+        compressImageMsg(*msg_img, *msg_img_compressed, {cv::IMWRITE_JPEG_QUALITY, jpeg_quality});
     }
     else if (format_type(cfg.pixelFormat) == FormatType::COMPRESSED) {
       // compressed image
@@ -633,6 +684,9 @@ CameraNode::onParameterChange(const std::vector<rclcpp::Parameter> &parameters)
 
         parameters_full[parameter.get_name()] = parameter.get_parameter_value();
       }
+    }
+    else if (!parameter.get_name().compare("jpeg_quality")) {
+      jpeg_quality = parameter.get_parameter_value().get<uint8_t>();
     }
   }
 


src/CameraNode.cpp
```diff
@@ -523,7 +536,41 @@ CameraNode::requestComplete(libcamera::Request *request)
 
       // compress to jpeg
       if (pub_image_compressed->get_subscription_count())
-        cv_bridge::toCvCopy(*msg_img)->toCompressedImageMsg(*msg_img_compressed);
+      {
+        std::vector<int> params;
+        params.reserve(2);
+        params.emplace_back(cv::IMWRITE_JPEG_QUALITY);
+        params.emplace_back(compression_quality);
```
christianrauch, line 543: Can you provide the parameter directly via a curled braces initialiser list, e.g. something like `{cv::IMWRITE_JPEG_QUALITY, compression_quality}`? This would be nicer to rad IMHO and reduce the code size.

src/CameraNode.cpp
```diff
@@ -523,7 +536,41 @@ CameraNode::requestComplete(libcamera::Request *request)
...
+
+        //only consider color jpeg compression
+        std::string targetFormat = "bgr8";
+        try
+        {
+          std::shared_ptr<CameraNode> tracked_object;
+          cv_bridge::CvImageConstPtr cv_ptr = cv_bridge::toCvShare(*msg_img,tracked_object, targetFormat);
+
+          // Compress image
+          if (cv::imencode(".jpg", cv_ptr->image, msg_img_compressed->data, params))
```
christianrauch, line 553: You are just setting the data here. But the image message needs a couple of more things to be set. Does this work like this? E.g. can you visualise the image in rqt without setting the `format`? Also, this definitely needs the header so it can be associated with a frame in time.

src/CameraNode.cpp
```diff
@@ -523,7 +536,41 @@ CameraNode::requestComplete(libcamera::Request *request)
...
+          {
+
+            float cRatio = (float)(cv_ptr->image.rows * cv_ptr->image.cols * cv_ptr->image.elemSize())
+                / (float)msg_img_compressed->data.size();
+            RCLCPP_DEBUG(get_logger(), "Compressed Image Transport - Codec: jpg, Compression Ratio: 1:%.2f (%lu bytes)", cRatio, msg_img_compressed->data.size());
```
christianrauch, line 558: Do we really need to compute this?

src/CameraNode.cpp
```diff
@@ -523,7 +536,41 @@ CameraNode::requestComplete(libcamera::Request *request)
...
+          }
+        }
+        catch (cv_bridge::Exception& e)
+        {
+          RCLCPP_ERROR(get_logger(), "%s", e.what());
+        }
+        catch (cv::Exception& e)
+        {
+          RCLCPP_ERROR(get_logger(), "%s", e.what());
+        }
```
christianrauch, line 572: Do we really need to handle all these errors here? What are the possible exceptions?

The original `toCompressedImageMsg` does not check the `[imencode](cv::imencode)` return value (but it probably should). But does `cv::imencode` throw exceptions? I don't see this from the documentation. Also `cv_bridge::toCvShare` doesn't seem to throw exceptions. If they do not throw exceptions, we don't need to handle them.

christianrauch (changes requested):
See the comments regarding the parameter, format conversion and error handling. For reference, this is the implementation of `toCompressedImageMsg` that you want to replace: https://github.com/ros-perception/vision_opencv/blob/066793a23e5d06d76c78ca3d69824a501c3554fd/cv_bridge/src/cv_bridge.cpp#L512-L535

It may be easier to just copy it and add the compression parameter to `cv::imencode`. Or better, send a PR to the `vision_opencv` repo with that additional parameter :-)

src/CameraNode.cpp
```diff
@@ -164,6 +167,16 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options) : Node("camera", opti
   // camera ID
   declare_parameter("camera", rclcpp::ParameterValue {}, param_descr_ro.set__dynamic_typing(true));
 
+  // jpeg compression quality
+  rcl_interfaces::msg::ParameterDescriptor param_descr_compression_role;
+  param_descr_compression_role.name = "jpeg_quality";
+  param_descr_compression_role.description = "quality of the jpeg compression, samller values lead to worst quality but lower bandwitdth";
+  param_descr_compression_role.additional_constraints = "integer value from 0 to 100";
+  param_descr_compression_role.type = rcl_interfaces::msg::ParameterType::PARAMETER_INTEGER;
+  param_descr_compression_role.read_only = false;
```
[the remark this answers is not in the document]
emrekuru97: Sure that's logical, i'll add this implemntation too. :)

src/CameraNode.cpp
```diff
@@ -523,7 +536,41 @@ CameraNode::requestComplete(libcamera::Request *request)
...
+        params.reserve(2);
+        params.emplace_back(cv::IMWRITE_JPEG_QUALITY);
+        params.emplace_back(compression_quality);
+
+        //only consider color jpeg compression
+        std::string targetFormat = "bgr8";
+        try
+        {
+          std::shared_ptr<CameraNode> tracked_object;
+          cv_bridge::CvImageConstPtr cv_ptr = cv_bridge::toCvShare(*msg_img,tracked_object, targetFormat);
```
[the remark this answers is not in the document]
emrekuru97: Regarding to the original implementation [here](https://github.com/christianrauch/camera_ros/blob/a23dfab7058b166dd7bfeb84df97a1bca7735561/src/CameraNode.cpp#L526C1-L526C82) 'toCvCopy' will do copying too, and also 'cv_bridge::toCvShare' uses the same. I could not see a solution for using 'cv::imencode' direclty with the original format.

src/CameraNode.cpp
emrekuru97: Sure it would look better, i will do this.

src/CameraNode.cpp
emrekuru97: I missed it i'll definitely add the header as well. But despite that the data was visualizable in rqt and rviz2 😲.

src/CameraNode.cpp
emrekuru97: I got this from [here](https://github.com/ros-perception/image_transport_plugins/blob/ef9442d8e92c14dcb5ed9023b9577e6d9e5cad39/compressed_image_transport/src/compressed_publisher.cpp#L216C9-L241C10). I thought it might be useful for debuggers, i can remove it.

src/CameraNode.cpp
emrekuru97: The original `compressedImagePublisher` handles these exception that was the reason i added it. As it can seen [here](https://github.com/ros-perception/image_transport_plugins/blob/ef9442d8e92c14dcb5ed9023b9577e6d9e5cad39/compressed_image_transport/src/compressed_publisher.cpp#L216C9-L241C10). I'll check the functions called in background maybe they throw exceptions.

christianrauch: I see. I thought you mainly wanted to re-implement `toCompressedImageMsg` with a quality option. But now I am wondering why `CompressedPublisher::publish` is doing all this error handling when `CvImage::toCompressedImageMsg` is not.

christianrauch:
From your comments, I can now see that you are essentially reimplementing `CompressedPublisher::publish` instead of `CvImage::toCompressedImageMsg`. The implementation of `CompressedPublisher::publish` is much more complex than `CvImage::toCompressedImageMsg`.

Do you think re-implemening `toCompressedImageMsg` with the compression quality option would be sufficient for you?

emrekuru97:
> From your comments, I can now see that you are essentially reimplementing `CompressedPublisher::publish` instead of `CvImage::toCompressedImageMsg`. The implementation of `CompressedPublisher::publish` is much more complex than `CvImage::toCompressedImageMsg`.
> 
> Do you think re-implemening `toCompressedImageMsg` with the compression quality option would be sufficient for you?
Yes it would be sufficient, if it's ok i will copy the [function](https://github.com/ros-perception/vision_opencv/blob/066793a23e5d06d76c78ca3d69824a501c3554fd/cv_bridge/src/cv_bridge.cpp#L512-L535) and make the necessary changes and update the pull request. But we still need to call `cv_bridge::toCvShare` to get a `CvImage`. Since the function `toCompressedImageMsg` works on `CvImage`. But i'll left the format blank so `cv_bridge::toCvShare` wont do any unnecessary converting.

src/CameraNode.cpp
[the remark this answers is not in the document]
emrekuru97: Implemented in new request.

src/CameraNode.cpp
```diff
@@ -164,6 +167,16 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options) : Node("camera", opti
   // camera ID
   declare_parameter("camera", rclcpp::ParameterValue {}, param_descr_ro.set__dynamic_typing(true));
 
+  // jpeg compression quality
+  rcl_interfaces::msg::ParameterDescriptor param_descr_compression_role;
+  param_descr_compression_role.name = "jpeg_quality";
+  param_descr_compression_role.description = "quality of the jpeg compression, samller values lead to worst quality but lower bandwitdth";
+  param_descr_compression_role.additional_constraints = "integer value from 0 to 100";
```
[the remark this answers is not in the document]
emrekuru97: Implented in new request.

src/CameraNode.cpp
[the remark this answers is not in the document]
emrekuru97: Keeping original format, with removing the `targetFormat` parameter.

src/CameraNode.cpp
emrekuru97: In new request provided the parameter directly via a curled braces.

src/CameraNode.cpp
emrekuru97: In new request desired fields are filled.

src/CameraNode.cpp
emrekuru97: Removed in new request.

src/CameraNode.cpp
emrekuru97: To let is simple in new reqest thes lines are removed.

emrekuru97:
I tryed to fix all requested changes, if necessary i can go to a new branch with only one commit(if no more issues are present) or let this as it is.

src/CameraNode.cpp
```diff
@@ -164,6 +170,21 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options) : Node("camera", opti
...
+  param_descr_compression_role.type = rcl_interfaces::msg::ParameterType::PARAMETER_INTEGER;
+  param_descr_compression_role.read_only = false;
+  rcl_interfaces::msg::IntegerRange jpeg_range;
+  jpeg_range.from_value = 1;
+  jpeg_range.to_value = 100;
```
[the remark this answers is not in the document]
emrekuru97: Removed comment, let it from 1 to 100.

src/CameraNode.cpp
```diff
@@ -164,6 +170,21 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options) : Node("camera", opti
   // camera ID
   declare_parameter("camera", rclcpp::ParameterValue {}, param_descr_ro.set__dynamic_typing(true));
 
+  // jpeg compression quality
+  rcl_interfaces::msg::ParameterDescriptor param_descr_compression_role;
+  param_descr_compression_role.name = "jpeg_quality";
```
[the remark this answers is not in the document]
emrekuru97: Changed param to jpeg_quality and comments to in whole file.

src/CameraNode.cpp
```diff
@@ -164,6 +170,21 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options) : Node("camera", opti
...
+  param_descr_compression_role.description = "quality of the jpeg compression, samller values lead to worst quality but lower bandwitdth";
```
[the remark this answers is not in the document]
emrekuru97: Changed comment to same as in `CompressedImagePublisher`.

src/CameraNode.cpp
```diff
@@ -164,6 +170,21 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options) : Node("camera", opti
...
+  param_descr_compression_role.type = rcl_interfaces::msg::ParameterType::PARAMETER_INTEGER;
+  param_descr_compression_role.read_only = false;
+  rcl_interfaces::msg::IntegerRange jpeg_range;
+  jpeg_range.from_value = 1;
+  jpeg_range.to_value = 100;
+  jpeg_range.step = 1;
+  param_descr_compression_role.integer_range = {jpeg_range};
+  //defult to 95
+  declare_parameter<int>("compression_quality", 95, param_descr_compression_role);
+  compression_quality =  get_parameter("compression_quality").as_int();
```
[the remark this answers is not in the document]
emrekuru97: Done in one linle.

src/CameraNode.cpp
```diff
@@ -108,6 +111,9 @@ class CameraNode : public rclcpp::Node
   // keep track of set parameters
   ParameterMap parameters_full;
   std::mutex parameters_lock;
+  // compression quality parameter
+  int compression_quality;
+  std::mutex compression_lock;
```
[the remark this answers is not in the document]
emrekuru97: Ok thx. Done this way.

src/CameraNode.cpp
```diff
@@ -523,7 +544,33 @@ CameraNode::requestComplete(libcamera::Request *request)
 
       // compress to jpeg
       if (pub_image_compressed->get_subscription_count())
-        cv_bridge::toCvCopy(*msg_img)->toCompressedImageMsg(*msg_img_compressed);
+      {
+        compression_lock.lock();
+        std::vector<int> params{cv::IMWRITE_JPEG_QUALITY,compression_quality};
+        compression_lock.unlock();
+        std::shared_ptr<CameraNode> tracked_object;
+        cv_bridge::CvImageConstPtr cv_ptr = cv_bridge::toCvShare(*msg_img,tracked_object);
```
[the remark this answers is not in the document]
emrekuru97: `toCvShare` does not use this parameter for execution, it attaches it to the returned pointer only as it can seen [here](https://github.com/ros-perception/vision_opencv/blob/066793a23e5d06d76c78ca3d69824a501c3554fd/cv_bridge/src/cv_bridge.cpp#L438C1-L456C2). The function `toCvShare` has two overloads the other overload does not need a `traked_object` but it requests `Image::ConstSharedPtr` which need a potential copy. Anyway I replaced it with `toCvCopy`, as can seen [here](https://github.com/ros-perception/vision_opencv/blob/066793a23e5d06d76c78ca3d69824a501c3554fd/cv_bridge/src/cv_bridge.cpp#L438C1-L456C2) `toCVShare` would use the same too, and i think for our case this will always true `(!encoding.empty() && source.encoding != encoding)`.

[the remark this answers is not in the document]
emrekuru97: At the beginning i tryed to replace `toCvShare` directly with `toCvCopy` but then i realized that it would copy always. If we look at the [implementation](https://github.com/ros-perception/vision_opencv/blob/066793a23e5d06d76c78ca3d69824a501c3554fd/cv_bridge/src/cv_bridge.cpp#L438C1-L456C2) the `tracked_object` is only attached to the returned `CvImage`.  `toCvShare` has a different overload without the `tracked_object` but this time it requests `Image::ConstSharedPtr` wich will potentially lead to a Copy in my opinion(i'm not sure). Let me now if there is a better and cheap as this solution for getting `CvImage` from `sensor_msgs::msg::Image`. I'll let it the same for this round. Thanks.

src/CameraNode.cpp
```diff
@@ -522,8 +537,33 @@ CameraNode::requestComplete(libcamera::Request *request)
...
+            temp = cv_bridge::cvtColor(tempThis, enc::BGRA8);
+          }
+          else {
+            temp = cv_bridge::cvtColor(tempThis, enc::BGR8);
+          }
+          image = temp->image;
+        }
+
+        msg_img_compressed->format = "jpg";
+        cv::imencode(".jpg", image, msg_img_compressed->data, params);
```
christianrauch, line 565: This is now a copy of [`CvImage::toCompressedImageMsg`](https://github.com/ros-perception/vision_opencv/blob/066793a23e5d06d76c78ca3d69824a501c3554fd/cv_bridge/src/cv_bridge.cpp#L512-L535). I suggest that we move this into a function and, more importantly, note where this code comes from and its license,

christianrauch (changes requested):
I am mostly fine with this now. The major thing is now to move this into a function and document its source and license.

src/CameraNode.cpp
```diff
@@ -60,12 +62,12 @@
 #include <unordered_map>
 #include <utility>
 #include <vector>
+namespace enc = sensor_msgs::image_encodings;
 namespace rclcpp
 {
 class NodeOptions;
 }
 
-
```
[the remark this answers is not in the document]
emrekuru97: I firstly formatted with wrong formatter, than with clang i think thats the rease. Fixed in new commit.

src/CameraNode.cpp
```diff
@@ -121,7 +125,6 @@ class CameraNode : public rclcpp::Node
 
 RCLCPP_COMPONENTS_REGISTER_NODE(camera::CameraNode)
 
-
```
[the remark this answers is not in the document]
emrekuru97: Fixed too.

src/CameraNode.cpp
```diff
@@ -522,8 +537,33 @@ CameraNode::requestComplete(libcamera::Request *request)
       memcpy(msg_img->data.data(), buffer_info[buffer].data, buffer_info[buffer].size);
 
       // compress to jpeg
-      if (pub_image_compressed->get_subscription_count())
-        cv_bridge::toCvCopy(*msg_img)->toCompressedImageMsg(*msg_img_compressed);
+      if (pub_image_compressed->get_subscription_count()) {
+        std::vector<int> params {cv::IMWRITE_JPEG_QUALITY, jpeg_quality};
```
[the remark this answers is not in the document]
emrekuru97: Done in new commit too.

src/CameraNode.cpp
emrekuru97: Moved it into a function and pushed last state. But i don't added any licenses befor as ı can see [here](https://github.com/ros-perception/vision_opencv/tree/humble?tab=License-1-ov-file) the code has two licences do we have to mentoin both or can we select wich one. According to this [discussion](https://stackoverflow.com/questions/1098608/what-does-it-mean-to-double-license) we are free to chose wich one to use. If so wich one should i use?  I think the BSD-3 license looks more friendly. Thx.

christianrauch: The file where the code is from lists `Software License Agreement (BSD License)` at the top.

src/CameraNode.cpp
```diff
@@ -522,8 +573,9 @@ CameraNode::requestComplete(libcamera::Request *request)
       memcpy(msg_img->data.data(), buffer_info[buffer].data, buffer_info[buffer].size);
 
       // compress to jpeg
-      if (pub_image_compressed->get_subscription_count())
-        cv_bridge::toCvCopy(*msg_img)->toCompressedImageMsg(*msg_img_compressed);
+      if (pub_image_compressed->get_subscription_count()) {
+        compressImageMsg(*msg_img, *msg_img_compressed);
+      }
```
christianrauch, line 578: Nitpick: Can you remove the curled braces here to follow the style of the code that you are replacing?

christianrauch (changes requested):
Thanks. The code looks much cleaner now. Missing is not only the source and license hint for the function that was copied (`compressImageMsg`) and a minor style change.

src/CameraNode.cpp
emrekuru97: Ohh yes, i will do it.

src/CameraNode.cpp
emrekuru97: Added the license information to the code too, but i'm not sure if this is the proper way.

src/CameraNode.cpp
```diff
@@ -481,6 +502,44 @@ CameraNode::declareParameters()
...
 }
 
+/*********************************************************************
+*  Copyright (c) 2011, Willow Garage, Inc.
+*  Copyright (c) 2015, Tal Regev.
+*  Copyright (c) 2018 Intel Corporation.
+*  All rights reserved.
+*  The function below compressImageMsg is taken from :https://github.com/ros-perception/vision_opencv
+*  And is slightly modified to meet the requirements of this code.
+**********************************************************************/
```
christianrauch, line 512: The most important information, that this part of the code is covered by the BSD License, is missing here. This section can also be shorter.

I would write something like:
The following function "compressImageMsg" is adapted from "CvImage::toCompressedImageMsg" (link to function source) and covered by the BSD-3-Clause licence.

IMHO the most important bits are where this comes from and under which licence it is. The res can be looked up in the original file.

src/CameraNode.cpp
```diff
@@ -481,6 +502,44 @@ CameraNode::declareParameters()
...
+    }
+    else {
+      temp = cv_bridge::cvtColor(tempThis, enc::BGR8);
+    }
+    image = temp->image;
+  }
+
+  destination.format = "jpg";
+  cv::imencode(".jpg", image, destination.data, {cv::IMWRITE_JPEG_QUALITY, jpeg_quality});
+}
```
christianrauch, line 541: To separate this "foreign" function better from the actual camera node, I suggest that this goes into its own object library or at least move it out of the `CameraNode` class into a normal function that takes the original parameters plus the compression quality.

christianrauch (changes requested):
I am still a bit worried about mixing the code with the copied implementation of `CvImage::toCompressedImageMsg`. Could you move this function either into its own library (cleaner) or into a stand-alone function outside of the `CameraNode` class (easier)?

src/CameraNode.cpp
emrekuru97: Ok changed.

src/CameraNode.cpp
emrekuru97: Moved the function out form the class. And added a extra parameter for the comressiopn quality.

emrekuru97:
I'll also try to get a PR in [original_repo](https://github.com/ros-perception/vision_opencv/tree/humble?tab=License-1-ov-file) with additional default parameter to the function. Or with a new signature. May be if they merge we can thane change this too.

src/CameraNode.cpp
```diff
@@ -481,6 +498,42 @@ CameraNode::declareParameters()
   set_parameters(parameters_init_list);
 }
 
+/*********************************************************************
+The following function "compressImageMsg" is adapted from "CvImage::toCompressedImageMsg" 
+(https://github.com/ros-perception/vision_opencv/blob/066793a23e5d06d76c78ca3d69824a501c3554fd/cv_bridge/src/cv_bridge.cpp#L512-L535) 
+and covered by the BSD-3-Clause licence.
+**********************************************************************/
```
christianrauch, line 505: Almost there. Can you use `//` for multi-line comments as in the other comment sections in this repo?

christianrauch (changes requested):

christianrauch:
> I'll also try to get a PR in [original_repo](https://github.com/ros-perception/vision_opencv/tree/humble?tab=License-1-ov-file) with additional default parameter to the function. Or with a new signature. May be if they merge we can thane change this too.

That would be awesome and remove the need to keep this function here.

src/CameraNode.cpp
emrekuru97: Done.

christianrauch (approved):

christianrauch (changes requested):
This looks good now. Final request: Can you squash all your commits into one and rebase/force-push to the PR? There are many back and forth changes that are better suited for a single commit.

emrekuru97:
> This looks good now. Final request: Can you squash all your commits into one and rebase/force-push to the PR? There are many back and forth changes that are better suited for a single commit.

Squashed the final code into the first commit, and changed the commit to "functionality for compression qualit".

src/CameraNode.cpp
```diff
@@ -481,6 +498,40 @@ CameraNode::declareParameters()
   set_parameters(parameters_init_list);
 }
 
+//The following function "compressImageMsg" is adapted from "CvImage::toCompressedImageMsg"
+//(https://github.com/ros-perception/vision_opencv/blob/066793a23e5d06d76c78ca3d69824a501c3554fd/cv_bridge/src/cv_bridge.cpp#L512-L535)
+//and covered by the BSD-3-Clause licence.
```
christianrauch, line 503: There should be a space between `//` and the comment text. See the other comments as an example. Sorry for being so picky about this but it just sticks out when the style differs from the rest of the code.

christianrauch (changes requested):

src/CameraNode.cpp
emrekuru97: Sure, i have done this.

emrekuru97:
Thanks :). Yesterday i also tryed to get a PR [here](https://github.com/ros-perception/vision_opencv/pull/521) in the original `cv_bridge` repo to the humble branch, but i'm not sure if that is the correct way since there are many request that are awaiting. Anyway if they will merge i'll remove the function and pass directly the parameter to the new version. I'll add the tag closes now.

Issue #37, opened 2024-04-24
IMX477 sensor configured as a sink


christian-nils:
Hi @christianrauch,

We have two IMX477 HQ cameras connected to a Raspberry Pi 5. One is configured as a sink while the other one is configured as a source.
Using rpicam-vid I can receive frames at 30FPS from both cameras but I am interested in using your implementation as it solves issues we have today with h264-encoded streams. Below are the commands I use with rpicam-vid

For the sink:
```shell
LIBCAMERA_RPI_CONFIG_FILE=/etc/rpi_apps.yaml && rpicam-vid -t 0 -n --camera 0 --codec libav --libav-format h264 --libav-video-codec libx264 --libav-video-codec-opts "tune=zerolatency;preset=ultrafast;g=30" --width 2028 --height 1080 --denoise off --lens-position 0.0 --hdr off --framerate 30 -o udp://127.0.0.1:5555 
```

For the source:
```shell
rpicam-vid -t 0 -n --camera 1 --codec libav --libav-format h264 --libav-video-codec libx264 --libav-video-codec-opts "tune=zerolatency;preset=ultrafast;g=30" --width 2028 --height 1080 --denoise off --lens-position 0.0 --hdr off --framerate 30 -o udp://127.0.0.1:5556
```

Using `camera_ros` with Raspberry Pi's libcamera, I can generate ROS2 compressed images, which is awesome but with some limitations. It only works with low FPS. For instance at 10 FPS, the following commands are used and works:

For the sink:
```shell
LIBCAMERA_RPI_CONFIG_FILE=/etc/rpi_apps.yaml && ros2 run  camera_ros camera_node --ros-args -p camera:=0 -p width:=2028 -p height:=1080 -p role:=still -r /camera/image_raw:=/left_cam/raw -r /camera/image_raw/compressed:=/left_cam/compressed -r /camera/camera_info:=/left_cam/camera_info -p FrameDurationLimits:="[100000,100000]" -p AeFlickerMode:="0" -p AeEnable:="false" -p HdrMode:="0" -p NoiseReductionMode:="0" -p format:="YUYV"
```

For the source:
```shell
[...]

PR #39, opened 2024-04-25
fix: the initial camera controls not used when starting camera

christian-nils:
Fixes issues including the one described in #37 where the sink camera is not aware of the targeted framerate causing problems with the autoexposure algorithm.

### DIFF
--- a/src/CameraNode.cpp
+++ b/src/CameraNode.cpp
@@ -379,8 +379,12 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options) : Node("camera", opti
   // register callback
   camera->requestCompleted.connect(this, &CameraNode::requestComplete);
 
+  libcamera::ControlList controls_ = camera->controls();
+  for (const auto &[id, value] : parameters)
+    controls_.set(id, value);
+
   // start camera and queue all requests
-  if (camera->start())
+  if (camera->start(&controls_))
     throw std::runtime_error("failed to start camera");
 
   for (std::unique_ptr<libcamera::Request> &request : requests)


christianrauch:
Thanks a lot for this PR. Is the issue related to https://github.com/christianrauch/camera_ros/issues/30?

Can you give an example of what did not work for you before this PR and how that changed with this PR?

christian-nils:
I don't think this will solve #30 (the bug report is also a bit confusing). I guess some of the listed stream formats are just not supported by the camera causing the issue reported.

For this PR, I can explain a bit more why it is relevant. In my case, I have two IMX477 sensors, one configured as a source and the other one as a sink using the following dtoverlay configuration:

```shell
dtoverlay=imx477,cam0,rotation=0,always-on,sync-sink
dtoverlay=imx477,rotation=0,always-on,sync-source
```

This means that the sink sensor will only capture frames when it receives a trigger on its XVS line.

If you don't add the expected framerate for the sink before starting the camera, you'll have issue with the AEC/AGC algorithm. After the 3rd frame of the startup phase (to initialize the algorithms I guess? those frames are dropped after https://github.com/raspberrypi/libcamera/blob/eb00c13d7c9f937732305d47af5b8ccf895e700f/src/libcamera/pipeline/rpi/pisp/data/example.yaml#L22), you will get an updated metadata with `Exposure=0`.

From the report:

```shell
[2:08:07.563255269] [673] DEBUG IPARPI ipa_base.cpp:1274 Metadata - Exposure: 985.83us Frame length: 2197 Line length: 15.17us Gain: 1
[2:08:07.563389046] [673] DEBUG IPARPI cam_helper.cpp:251 Metadata updated - Exposure: 0.00us Frame length: 0 Line length: 0.00us Gain: 1 Temperature: 0
```

This will happen for every following frames.

When using the fix, you set the controls before the camera as explained here (https://libcamera.org/api-html/classlibcamera_1_1Camera.html#a5c033a72ac66fbdd2c2bb9f934389dcf). Which leads to a fixing the error reported above:

```shell
[2:19:04.777075042] [862] DEBUG IPARPI ipa_base.cpp:1274 Metadata - Exposure: 985.83us Frame length: 2175 Line length: 15.17us Gain: 1
[2:19:04.777131041] [862] DEBUG IPARPI cam_helper.cpp:251 Metadata updated - Exposure: 985.83us Frame length: 2175 Line length: 15.17us Gain: 1 Temperature: 27
```

So, in short, we get the following before the fix (obstructing the first frames to get a high exposure which bugs the successive capture): 
[bad.webm](https://github.com/christianrauch/camera_ros/assets/2658014/ef5d6d8f-ff72-46d2-94f1-6bf26f2063eb)

we get the following after applying the patch
[good.webm](https://github.com/christianrauch/camera_ros/assets/2658014/7026a785-37da-414b-9708-950e3fe3f623)

Please let me know if that addresses your questions.

src/CameraNode.cpp
```diff
@@ -379,8 +379,12 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options) : Node("camera", opti
   // register callback
   camera->requestCompleted.connect(this, &CameraNode::requestComplete);
 
+  libcamera::ControlList controls_ = camera->controls();
+  for (const auto &[id, value] : parameters)
+    controls_.set(id, value);
```
christianrauch, line 384: This is a bit fragile as it depends on the `declareParameters` call before. That parameter declaration will be followed immediately by calls to `onParameterChange` for every individual entry in the "controls" which in turn sets the `parameters`. Additionally, this is duplicated in `requestComplete` which also reads from `parameters` and sets controls (but for the requests).

Is there a nicer way to do this? Let's say instead of managing the `parameters` of type `unordered_map` could we use a `ControlList` directly?

christian-nils: Do you mean stop using parameters as a `unordered_map` in the whole package but using it as a `ControlList`? If so, I think this is how they do in the `rpicam-apps`

I am not sure I'll be able to make a new proposal over the weekend, will try to do that next week. Have a great one ! :)

christianrauch: Ok. Nevermind. I will merge this "as is" if it helps you to go in with your project.

christianrauch (approved):

src/CameraNode.cpp
christian-nils: Thank you very much!

PR #59, opened 2024-09-09
Stereo threaded

Ingaiza:
The initial node has some lag/delay between the left and right images even though the image info has the same timestamp, so this is the first attempt to solve that issue using threads, The output still has some little delay, still looking for other solutions. Looking to try ROS2 MultiThreadedExecutors as it may turn out as a better option.


christianrauch:
What are you trying to do? This is the second PR to this that you open and close immediately. Are you trying to add this `stereonode` code to the repo?

PR #60, opened 2024-09-10
Stereo executors

Ingaiza:
Used MultiThreadedExecutors to run the left and right image publishing separately


Ingaiza:
My apologies for that

PR #90, opened 2025-02-12
Add ROS parameter for setting camera info file URL

samyarsadat:
As the title suggests, this PR adds a new ROS parameter (`ci_file_url`) for setting the URL of the camera calibration info file used by CameraInfoManager.

### DIFF
--- a/src/CameraNode.cpp
+++ b/src/CameraNode.cpp
@@ -220,6 +220,11 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options) : Node("camera", opti
   declare_parameter<int64_t>("width", {}, param_descr_ro);
   declare_parameter<int64_t>("height", {}, param_descr_ro);
 
+  // camera info file url
+  rcl_interfaces::msg::ParameterDescriptor param_descr_camera_info_url;
+  param_descr_camera_info_url.description = "camera calibration info file url";
+  param_descr_camera_info_url.read_only = true;
+
   // camera ID
   declare_parameter("camera", rclcpp::ParameterValue {}, param_descr_ro.set__dynamic_typing(true));
 
@@ -398,6 +403,15 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options) : Node("camera", opti
 
   if (!cim.setCameraName(cname))
     throw std::runtime_error("camera name must only contain alphanumeric characters");
+
+  const std::string &camera_info_url = declare_parameter<std::string>(
+    "camera_info_url", {}, param_descr_camera_info_url);
+  if (!cim.loadCameraInfo(camera_info_url)) {
+    if (!camera_info_url.empty()) {
+      RCLCPP_WARN_STREAM(get_logger(), "failed to load camera calibration info from provided URL, using default URL");
+      cim.loadCameraInfo({});
+    }
+  }
 
   declareParameters();
 


src/CameraNode.cpp
```diff
@@ -220,6 +220,12 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options) : Node("camera", opti
   declare_parameter<int64_t>("width", {}, param_descr_ro);
   declare_parameter<int64_t>("height", {}, param_descr_ro);
 
+  // camera info file url
+  rcl_interfaces::msg::ParameterDescriptor param_descr_ci_file_url;
+  param_descr_ci_file_url.description = "camera calibration info file url";
+  param_descr_ci_file_url.read_only = true;
+  declare_parameter<std::string>("ci_file_url", {}, param_descr_ci_file_url);
```
christianrauch, line 227: The [`camera_info_manager` documentation](https://github.com/ros-perception/image_common/blob/b021202a42d7bc3ff3215237ded251a59c282da5/camera_info_manager/include/camera_info_manager/camera_info_manager.hpp#L99-L104) suggests using the parameter name `camera_info_url`:

> The location for getting and saving calibration data is expressed by Uniform Resource Locator.  The driver defines a URL via the CameraInfoManager::CameraInfoManager constructor or the loadCameraInfo() method.  Many drivers provide a ~camera_info_url parameter so users may customize this URL, but that is handled outside this class.

src/CameraNode.cpp
```diff
@@ -396,6 +402,17 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options) : Node("camera", opti
   std::replace_if(
     cname.begin(), cname.end(), [](const char &x) { return !std::isalnum(x); }, '_');
 
+  std::string ci_file_url_param = get_parameter("ci_file_url").as_string();
```
christianrauch, line 405: You can directly get the value from `declare_parameter`, e.g. `const std::string &ci_file_url = declare_parameter()`.

src/CameraNode.cpp
```diff
@@ -396,6 +402,17 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options) : Node("camera", opti
...
+  if (!ci_file_url_param.empty()) {
+    if (cim.validateURL(ci_file_url_param)) {
```
christianrauch, line 407: Is this dedicated `validateURL` required? `loadCameraInfo` also returns a `bool` and according to the [`loadCameraInfo` documentation](https://github.com/ros-perception/image_common/blob/b021202a42d7bc3ff3215237ded251a59c282da5/camera_info_manager/src/camera_info_manager.cpp#L312-L335):

> @return true if new URL contains calibration data.

I think we can skip `validateURL` here if we want to apply the values anyway.

src/CameraNode.cpp
```diff
@@ -396,6 +402,17 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options) : Node("camera", opti
...
+      RCLCPP_INFO_STREAM(get_logger(), "loading camera calibration info from \"" << ci_file_url_param << "\"");
+      cim.loadCameraInfo(ci_file_url_param);
+    }
+    else {
+      throw std::runtime_error("invalid camera calibration info file url");
```
christianrauch, line 412: I think I would rather fall back to the default URL (`file://${ROS_HOME}/camera_info/${NAME}.yaml`) and show an error here instead of throwing an unhandled exception. What do you think?

christianrauch (changes requested):

src/CameraNode.cpp
samyarsadat: Agreed, I'll just use the return value of `loadCameraInfo`.

src/CameraNode.cpp
samyarsadat: Yeah, I think that would be much better. Should've done it that way in the first place. I'll change it.

samyarsadat:
All done.

src/CameraNode.cpp
```diff
@@ -399,6 +404,13 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options) : Node("camera", opti
   if (!cim.setCameraName(cname))
     throw std::runtime_error("camera name must only contain alphanumeric characters");
 
+  std::string ci_file_url_param = declare_parameter<std::string>("camera_info_url", {},
+                                                                 param_descr_ci_file_url);
```
christianrauch, line 408: Is this auto-formatting? Can you move the `"camera_info_url"` and the rest into the second line? Also, make `ci_file_url_param` a const reference to save a copy and maybe call the variable `camera_info_url` like the parameter name to fit in with the rest of the code style.

src/CameraNode.cpp
```diff
@@ -399,6 +404,13 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options) : Node("camera", opti
...
+  if (!ci_file_url_param.empty() && !cim.loadCameraInfo(ci_file_url_param)) {
+    RCLCPP_WARN_STREAM(get_logger(), "failed to load camera calibration info from provided URL, using default URL");
+    cim.loadCameraInfo("file://${ROS_HOME}/camera_info/${NAME}.yaml");
```
christianrauch, line 411: Can we get the default URL directly from the camera manager instead of copying the string?

christianrauch (changes requested):

src/CameraNode.cpp
samyarsadat: Haha, no, manual formatting. I'll fix it.

src/CameraNode.cpp
samyarsadat: I looked into it, and there doesn't seem to be a way for getting the default URL. It's [defined as a `const std::string`](https://github.com/ros-perception/image_common/blob/43699c5660ae5aeaadccc83ad12a3ce779d39fd0/camera_info_manager/src/camera_info_manager.cpp#L68C1-L69C75) in `camera_info_manager.cpp`, but we can't access it. 

I was thinking maybe we could do something like before:
```cpp
if (!ci_file_url_param.empty()) {
  if (cim.validateURL(ci_file_url_param)) {
    cim.loadCameraInfo(ci_file_url_param);
  }
  else {
    RCLCPP_WARN_STREAM(get_logger(), "invalid camera calibration info file URL, using default URL");
  }
}
```
This way if the URL is invalid, `loadCameraInfo()` is never called, and so the calibration file is loaded from the default URL when `getCameraInfo()` is called later on.

There is one main difference between this and checking the return value of `loadCameraInfo()` though: if there are any parsing errors, or the URL fails to resolve, the program won't fall back to the default URL, as we're only checking the validity of the URL itself.

christianrauch: Hm, I had a look at the code myself.

`loadCameraInfo` overrides the internal storage of the URL (`url_`) and the flag indicating that a camera info was loaded (`loaded_cam_info_`) before `loadCalibration` is called. If this fails, the values are not reset. That is not great since consecutive calls to `getCameraInfo` will return the empty/default `cam_info_`.

But looking further, you can see that calling `loadCameraInfo` with an empty string for the URL will eventually call `loadCalibration(default_camera_info_url, cname);` in `loadCalibration`.

So, I propose the following procedure:
1. declare parameter `"camera_info_url"` with an empty default value
2. call `loadCameraInfo` with the parameter value and check return value
    - if return is true, then either the provided URL is correct and contains camera info data or if no URL was provided, the default URL contains valid data, DONE
    - if false and the parameter is empty, then nothing was provided and there is no valid data at the default URL, DONE
    - if false and the parameter is not empty, show an error message that there is no valid calibration data and the URL and tell the user that there will be an attempt to real from the default URL, call `loadCameraInfo` again with an empty URL (`{}`) to load from the default, DONE

We do not need to check the last `loadCameraInfo` call with the default URL since it will show an error anyway if there is no data at the default URL and we cannot do anything about it.

samyarsadat: I see... I assumed that calling `loadCameraInfo` once with a URL permanantly overrode the default URL. Given this new information, your proposal seems like the best option.

samyarsadat: I've implemented yout suggestion and tested it with my setup. Everything seems to work as intended.

christianrauch (approved):

PR #95, opened 2025-02-15
Consider control extent when clamping control values

samyarsadat:
I've extracted the `clamp` fix from my other PR. As I mentioned there:

> I also had to make some minor modifications to the `clamp` and `clamp_array` functions. This is because there can be cases in which _scalar_ `min`, `max`, and `default` values for a _span_ control are passed to `clamp`. In such cases, `clamp` would return a scalar value, which would then be set as the default for the aforementioned _span_ control (in `CameraNode::declare_parameters()`). This exact scenario occurs with `AfWindows`. `AfWindows` is a span type, but its `min`, `max`, and `default` values are scalar. 
>
> _(Note: when referring to span controls above, I'm referring to dynamic ones, not fixed-length ones)_
>
> I fixed this issue by passing the extent of the control to `clamp` and using that to determine whether the control is scalar or not instead of using `value.isArray()`. If the control has dynamic extent, a span `ControlValue` with one element (clamped `value`) is returned.

More clearly: libcamera (at least the RPi fork, as that's what I'm testing with) sometimes provides scalar default values for span controls. In those cases, the current `clamp()` also returns a scalar value which is then set as the default for that control. However, I thought it might be better (for controls with dynamic extent) for `clamp` to return a span with a single element (clamped scalar default) instead of just returning a scalar value.

To achieve this, I did have to make it so that the extent of the control is also passed to `clamp`, which might not be desirable; but given that `clamp` is only used when setting default control values, I don't think this is going to be an issue.

### DIFF
--- a/src/CameraNode.cpp
+++ b/src/CameraNode.cpp
@@ -580,7 +580,7 @@ CameraNode::declareParameters()
     // clamp default ControlValue to min/max range and cast ParameterValue
     rclcpp::ParameterValue value;
     try {
-      value = cv_to_pv(clamp(info.def(), info.min(), info.max()));
+      value = cv_to_pv(clamp(info.def(), info.min(), info.max(), extent));
     }
     catch (const invalid_conversion &e) {
       RCLCPP_ERROR_STREAM(get_logger(), "unsupported control '"
--- a/src/clamp.cpp
+++ b/src/clamp.cpp
@@ -15,7 +15,7 @@
 
 #define CASE_CLAMP(T)             \
   case libcamera::ControlType##T: \
-    return clamp<ControlTypeMap<libcamera::ControlType##T>::type>(value, min, max);
+    return clamp<ControlTypeMap<libcamera::ControlType##T>::type>(value, min, max, extent);
 
 #define CASE_NONE(T)              \
   case libcamera::ControlType##T: \
@@ -104,16 +104,27 @@ clamp(const CTPoint &val, const CTPoint &lo, const CTPoint &hi)
 template<typename T>
 libcamera::ControlValue
 clamp_array(const libcamera::ControlValue &value, const libcamera::ControlValue &min,
-            const libcamera::ControlValue &max)
+            const libcamera::ControlValue &max, size_t extent)
 {
-  const libcamera::Span<const T> v = value.get<libcamera::Span<const T>>();
-  const libcamera::Span<const T> a = min.get<libcamera::Span<const T>>();
-  const libcamera::Span<const T> b = max.get<libcamera::Span<const T>>();
+  std::vector<T> vclamp;
 
-  std::vector<T> vclamp(v.size());
+  if (!value.isArray() && extent != 0) {
+    if (extent == libcamera::dynamic_extent) {
+      vclamp.push_back(std::clamp(value.get<const T>(), min.get<const T>(), max.get<const T>()));
+    }
+    else {
+      throw std::runtime_error("cannot clamp scalar value for fixed-extent span control");
+    }
+  }
+  else {
+    const libcamera::Span<const T> v = value.get<libcamera::Span<const T>>();
+    const libcamera::Span<const T> a = min.get<libcamera::Span<const T>>();
+    const libcamera::Span<const T> b = max.get<libcamera::Span<const T>>();
 
-  for (size_t i = 0; i < v.size(); i++)
-    vclamp[i] = std::clamp(v[i], a[i], b[i]);
+    for (size_t i = 0; i < v.size(); ++i) {
+      vclamp.push_back(std::clamp(v[i], a[i], b[i]));
+    }
+  }
 
   return libcamera::ControlValue(libcamera::Span<const T>(vclamp));
 }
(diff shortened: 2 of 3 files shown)


christianrauch:
> However, I thought it might be better (for controls with dynamic extent) for `clamp` to return a span with a single element (clamped scalar default) instead of just returning a scalar value.

Can you give an example where this fixes an issue, or why would this be useful otherwise? Semantically, a scalar and an array with a single value are the same.

If your control is a Span with more than one element (there are no 1-element-span controls in libcamera yet) and has a scalar default value, it is unclear how this should be interpreted. Therefore, in https://github.com/christianrauch/camera_ros/pull/89/commits/7f216a348bb298a591cec873d513e39d38851244 I am checking for this and ignore the control.

samyarsadat:
> Can you give an example where this fixes an issue, or why would this be useful otherwise? Semantically, a scalar and an array with a single value are the same.

When testing my other PR, I notcied that a scalar default was being set for `AfWindows`, so that's why I tried to fix it. But if, as you say, an array with a single value and a scalar value are equivalent, then it really doesn't matter and this PR can be discarded.

I must apologize, I don't have much experience with libcamera...

christianrauch:
The problem I was fixing with ignoring scalar default values for Span controls manifest for example with the [`FrameDurationLimits` control in the Raspberry Pi IPA](https://github.com/raspberrypi/libcamera/blob/29156679717bec7cc4784aeba3548807f2c27fca/src/ipa/rpi/common/ipa_base.cpp#L1131-L1135) here:
```C++
case controls::FRAME_DURATION_LIMITS: {
    auto frameDurations = ctrl.second.get<Span<const int64_t>>();
    applyFrameDurations(frameDurations[0] * 1.0us, frameDurations[1] * 1.0us);
    break;
}
```

`FrameDurationLimits` is of type `Control<Span<const int64_t, 2>>` but since https://github.com/raspberrypi/libcamera/commit/8cebd777cae428daf415998cc588fe60c6de0d66, its default `ControlValue` is set to the scalar of type `int64_t`. Hence, `ControlValue::isArray()` will return false for the default value.

Taking that default scalar value and setting it on the control will later trigger an assert when you try to `ControlValue::get<Span<const int64_t>>()` since this can only be used with array `ControlValue`s where `isArray()` returns true.

Now, converting that scaler `ControlValue`, with `isArray()` false into an 1-element array `ControlValue`, with `isArray()` true, will avoid triggering that assert. But you will still get only a single element, such that accessing the second element of `frameDurations` (`frameDurations[1]`) will either crash the process in the best case, or read garbage in the worst case.

If you see a use case where converting the scalar `ControlValue` into an 1-element `ControlValue` is useful, I can have a look at the PR. But otherwise, I don't see how this would be used.

samyarsadat:
In the case of controls like `FrameDurationLimits`, my PR won't make a difference as `FrameDurationLimits` has fixed length (i.e. its extent isn't dynamic). This PR makes it so that single value spans are returned _only_ when the the size/extent of the `ControlValue` is dynamic. If `clamp` were to be called with a scalar default value for a fixed-size control (which won't happen because of 7f216a3), it would throw an exception.

Here you can see that (excerpt from PR):
```cpp
if (!value.isArray() && extent != 0) {
  if (extent == libcamera::dynamic_extent) {
    vclamp.push_back(std::clamp(value.get<const T>(), min.get<const T>(), max.get<const T>()));
  }
  else {
    throw std::runtime_error("cannot clamp scalar value for fixed-extent span control");
  }
}
```

This PR can be useful in cases where a scalar default value is provided for a dynamic extent `ControlValue`. 7f216a3 doesn't prevent this from happening.

There something I am unsure about, however. Is it okay to set a scalar value for a `ControlValue` with dynamic extent? I presume the issue you mentioned with the assertion when calling `ControlValue::get<Span<const T>>()` would apply here as well, no?

I only saw this problem with the `AfWindows` control (when testing my other PR), which currently isn't supported by `camera_ros`. However, if support for nested `Rectangle`s (and hence `AfWindows`) is added, then it would become a problem potentially.

christianrauch:
> There something I am unsure about, however. Is it okay to set a scalar value for a `ControlValue` with dynamic extent? I presume the issue you mentioned with the assertion when calling `ControlValue::get<Span<const T>>()` would apply here as well, no?

If you `get()` the same type as what the `ControlValue` actually stores, no assertion will fail. But if you have additional assumptions about the size of the spans, such as reading 3 values from a span with only 2 elements, something else will certainly fail. There are no checks in libcamera that prevent you from setting a 3-element array `ControlValue` on a 3-element span `Control<>`. That is why these checks are done in `camera_ros`. The same goes for min/max/def `ControlValue`s with different types, which is possible in libcamera but can cause weird behaviour.

> I only saw this problem with the `AfWindows` control (when testing my other PR), which currently isn't supported by `camera_ros`. However, if support for nested `Rectangle`s (and hence `AfWindows`) is added, then it would become a problem potentially.

Since `AfWindows` is of type `Span<const Rectangle>`, I assume you are supposed to set one or many `Rectangle`. At least, this is what I can conclude from its sole usage in [`ipaAfControls`](https://github.com/raspberrypi/libcamera/blob/29156679717bec7cc4784aeba3548807f2c27fca/src/ipa/rpi/common/ipa_base.cpp#L101). But since the values are nowhere read, you could set anything at the moment. So once `AfWindows` is actively used, we could consider converting the default value. But at the moment, this will have no effect.

samyarsadat:
> If you `get()` the same type as what the `ControlValue` actually stores, no assertion will fail. But if you have additional assumptions about the size of the spans, such as reading 3 values from a span with only 2 elements, something else will certainly fail. There are no checks in libcamera that prevent you from setting a 3-element array `ControlValue` on a 3-element span `Control<>`. That is why these checks are done in `camera_ros`. The same goes for min/max/def `ControlValue`s with different types, which is possible in libcamera but can cause weird behaviour.

Wouldn't there be an assumption for a dynamic-extent span control to be of type `Span<const T>`, and wouldn't the assertion fail if one tried to `get()` it as that type if the control has been set with a scalar value?

> Since `AfWindows` is of type `Span<const Rectangle>`, I assume you are supposed to set one or many `Rectangle`. At least, this is what I can conclude from its sole usage in [`ipaAfControls`](https://github.com/raspberrypi/libcamera/blob/29156679717bec7cc4784aeba3548807f2c27fca/src/ipa/rpi/common/ipa_base.cpp#L101). But since the values are nowhere read, you could set anything at the moment. So once `AfWindows` is actively used, we could consider converting the default value. But at the moment, this will have no effect.

I see. So yes, in that case this PR wouldn't really change anything for `AfWindows`, and there aren't any other controls in a similar situation that I know of. We could revisit this when `AfWindows` gets properly implemented, if you think that is more appropiate.

christianrauch:
> Wouldn't there be an assumption for a dynamic-extent span control to be of type `Span<const T>`, and wouldn't the assertion fail if one tried to `get()` it as that type if the control has been set with a scalar value?

Yes. This is exactly what I explained in https://github.com/christianrauch/camera_ros/pull/95#issuecomment-2660931040. The `FrameDurationLimits` is a span control, but its default is a scalar value. So the assertion in `get` fails.

> I see. So yes, in that case this PR wouldn't really change anything for `AfWindows`, and there aren't any other controls in a similar situation that I know of. We could revisit this when `AfWindows` gets properly implemented, if you think that is more appropiate.

Unless we can check the code to see how values are read, there is no point of converting the type IMHO. It's just a code change and additional processing without an effect.

samyarsadat:
> Yes. This is exactly what I explained in [#95 (comment)](https://github.com/christianrauch/camera_ros/pull/95#issuecomment-2660931040). The `FrameDurationLimits` is a span control, but its default is a scalar value. So the assertion in `get` fails.

I see, so it applies to all span controls, as I assumed. 

> Unless we can check the code to see how values are read, there is no point of converting the type IMHO. It's just a code change and additional processing without an effect.

Very well then. We can shelve this for later. I'll keep the PR branch.

PR #100, opened 2025-02-27
Enable selection of sensor format to avoid a cropped FoV

### LINKED ISSUE #99: RPi Camera v2: Cropped image at low resolution
I am trying to use camera_ros to get a full FoV picture at a low resolution (640x480 or even 160x120) on Raspberry Pi 4, using the Camera Module v2, on Ubuntu 22.04/24.04.

Using the old legacy camera support then 640x480 was the default, but with libcamera (and hence camera_ros) it picks a sensor mode that leads to cropping, rather than scaling down.

Is there a way to get the sensor input for the lowest full FoV resolution (1640x1232), but then use the GPU to scale it to the desired lower resolution to avoid 'overloading' the ROS nodes having to serialize/deserialize the larger image?

Reference:
https://picamera.readthedocs.io/en/release-1.13/fov.html#sensor-modes

larsll:
Work to resolve #99 

It enables the raw role, and adjusts the size of it to match the desired input, this avoids cropping where libcamera by default selects a sensor mode that is not having the full FoV. This is a problem with RPi Camera V2 (see https://picamera.readthedocs.io/en/release-1.13/fov.html#sensor-modes).

**Example:**
```
ros2 run camera_ros camera_node --ros-args -p format:=BGR888 -p height:=120 -p width:=160 -p mode_width:=1640 -p mode_height:=1232
```
**Output:**
```
[INFO] [1740686517.425357897] [camera]: 1640x1232
[INFO] [1740686517.425436891] [camera]: Mode configuration:1640x1232-SBGGR12_CSI2P
[WARN] [1740686517.425989703] [camera]: stream configuration adjusted from "160x120-BGR888" to "160x120-BGR888"
[0:46:50.002516247] [3249]  INFO Camera camera.cpp:1033 configuring streams: (0) 160x120-BGR888 (1) 1640x1232-SBGGR10_CSI2P
[0:46:50.003349197] [3259]  INFO RPI vc4.cpp:512 Sensor: /base/soc/i2c0mux/i2c@1/imx219@10 - Selected sensor format: 1640x1232-SBGGR10_1X10 - Selected unicam format: 1640x1232-pBAA
[INFO] [1740686517.429272329] [camera]: camera "/base/soc/i2c0mux/i2c@1/imx219@10" configured with 160x120-BGR888 stream
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,7 @@ The camera stream is configured once when the node starts via the following stat
 | `role`            | `string`              | configures the camera with a `StreamRole` (possible choices: `raw`, `still`, `video`, `viewfinder`) [default: `viewfinder`] |
 | `format`          | `string`              | a `PixelFormat` that is supported by the camera [default: auto]                                                             |
 | `width`, `height` | `integer`             | desired image resolution [default: auto]                                                                                    |
+| `sensor_mode`     | `string`              | desired raw sensor format resolution (format: `width:height`) [default: auto]                                               |
 
 
 The configuration is done in the following order:
@@ -130,12 +131,21 @@ The configuration is done in the following order:
 2. configure camera stream via `role`
 3. set the pixel format for the stream via `format`
 4. set the image resolution for the stream via `width` and `height`
+5. set the sensor mode resolution for the raw feed from camera to GPU
 
 Each stream role only supports a discrete set of data stream configurations as a combination of the image resolution and pixel format. The selected stream configuration is validated at the end of this sequence and adjusted to the closest valid stream configuration.
 
 By default, the node will select the first available camera and configures it with the default pixel format and image resolution. If a parameter has not been set, the node will print the available options and inform the user about the default choice.
 
 The node avoids memory copies of the image data by directly mapping from a camera pixel format to a ROS image format, with the exception of converting between "raw" and "compressed" image formats when requested by the user. As an effect, not all pixel formats that are supported by the camera may be supported by the ROS image message. Hence, the options for `format` are limited to pixel formats that are supported by the camera and the raw ROS image message.
+
+### Sensor Modes and Cropping
+
+Most camera modules provide different sensor modes, some of which provide a smaller resolution version of the full picture, whereas others have a limited field-of-view, e.g. by allowing for the selection of a 3:2 image from a 16:9 sensor, a more native 'digital zoom' effect by cropping the picture etc. The benefit of this is to perform compute intensive tasks at the source (the sensor), rather than downstream in the GPU or in application code. Unless we specify the sensor mode we want, libcamera will automatically select one. Unfortunately, it is not guaranteed that the selected mode will be a full sensor mode, leading to a potential cropping of the picture. 
+
+Example: When configuring a 640x480 output stream on a Raspberry Pi Camera Module v2, libcamera will automatically choose a heavily cropped 640x480 sensor mode, leading to a digital zoom of ~2.5x.
+
+See the [PiCamera2 Documentation, section 4.2.2.3](https://datasheets.raspberrypi.com/camera/picamera2-manual.pdf) for a detailed explanation. The [PiCamera Documentation](https://picamera.readthedocs.io/en/release-1.13/fov.html#sensor-modes) has a good visualization of the issue. Read the documentation of your camera to see what modes are available, and which (if any) raw formats are cropped.
 
 ### Dynamic Frame Configuration (controls)
 
--- a/src/CameraNode.cpp
+++ b/src/CameraNode.cpp
@@ -47,6 +47,7 @@
 #include <rclcpp/publisher.hpp>
 #include <rclcpp/time.hpp>
 #include <rclcpp_components/register_node_macro.hpp>
+#include <regex>
 #include <sensor_msgs/image_encodings.hpp>
 #include <sensor_msgs/msg/camera_info.hpp>
 #include <sensor_msgs/msg/compressed_image.hpp>
@@ -150,6 +151,35 @@ get_role(const std::string &role)
   }
 }
 
+libcamera::Size
+get_sensor_format(const std::string &format_str)
+{
+  if (format_str.empty()) {
+    return {};
+  }
+
+  const std::regex pattern(R"((\d+):(\d+))");
+  std::smatch match;
+
+  if (std::regex_match(format_str, match, pattern)) {
+    try {
+      const int width = std::stoi(match[1].str());
+      const int height = std::stoi(match[2].str());
+
+      return libcamera::Size {static_cast<unsigned int>(width), static_cast<unsigned int>(height)};
+    }
+    catch (const std::out_of_range &) {
+      throw std::runtime_error("Invalid sensor_mode. Width or height exceeds maximum value of " +
+                               std::to_string(std::numeric_limits<int>::max()));
+    }
+    catch (const std::invalid_argument &) {
+      // Unexpected - throw exception below
+    }
+  }
+
+  // Throw exception as it was not possible to parse the format string
+  throw std::runtime_error("Invalid sensor_mode. Expected [width]:[height] but got " + format_str);
+}
 
 // The following function "compressImageMsg" is adapted from "CvImage::toCompressedImageMsg"
 // (https://github.com/ros-perception/vision_opencv/blob/066793a23e5d06d76c78ca3d69824a501c3554fd/cv_bridge/src/cv_bridge.cpp#L512-L535)
@@ -220,6 +250,13 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options)
   const uint32_t w = declare_parameter<int64_t>("width", {}, param_descr_ro);
   const uint32_t h = declare_parameter<int64_t>("height", {}, param_descr_ro);
   const libcamera::Size size {w, h};
+
+  // Raw format dimensions
+  rcl_interfaces::msg::ParameterDescriptor param_descr_sensor_mode;
+  param_descr_sensor_mode.description = "raw mode of the sensor";
+  param_descr_sensor_mode.additional_constraints = "string in format [width]:[height]";
+  param_descr_sensor_mode.read_only = true;
+  const libcamera::Size sensor_size = get_sensor_format(declare_parameter<std::string>("sensor_mode", {}, param_descr_sensor_mode));
 
   // camera info file url
   rcl_interfaces::msg::ParameterDescriptor param_descr_camera_info_url;
@@ -299,14 +336,20 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options)
   if (camera->acquire())
     throw std::runtime_error("failed to acquire camera");
 
+  std::vector<libcamera::StreamRole> roles {role};
+
+  // Add the RAW role if the sensor_size is defined
+  if (!sensor_size.isNull() && role != libcamera::StreamRole::Raw) {
+    roles.push_back(libcamera::StreamRole::Raw);
+  }
+
   // configure camera stream
   std::unique_ptr<libcamera::CameraConfiguration> cfg =
-    camera->generateConfiguration({role});
+    camera->generateConfiguration(roles);
 
-  if (!cfg)
-    throw std::runtime_error("failed to generate configuration");
+  if (!cfg || cfg->size() != roles.size())
+    throw std::runtime_error("failed to generate configuration for all roles");
 
-  assert(cfg->size() == 1);
   libcamera::StreamConfiguration &scfg = cfg->at(0);
 
   // list all camera formats, including those not supported by the ROS message
@@ -357,6 +400,12 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options)
     scfg.size = size;
   }
 
+  if (!sensor_size.isNull() && role != libcamera::StreamRole::Raw) {
+    libcamera::StreamConfiguration &modecfg = cfg->at(1);
+    modecfg.size = sensor_size;
+    RCLCPP_INFO_STREAM(get_logger(), "Sensor mode configuration: " << modecfg.toString());
+  }
+
   // store selected stream configuration
   const libcamera::StreamConfiguration selected_scfg = scfg;
 
@@ -392,6 +441,8 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options)
   const std::optional<std::string> model = props.get(libcamera::properties::Model);
   if (model)
     cname = model.value() + '_' + cname;
+  if (!sensor_size.isNull() && role != libcamera::StreamRole::Raw)
+    cname = cname + '_' + cfg->at(1).toString();
 
   // clean camera name of non-alphanumeric characters
   cname.erase(


src/CameraNode.cpp
```diff
@@ -357,6 +368,19 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options)
     scfg.size = size;
   }
 
+  if (!mode_size.isNull()) {
+    RCLCPP_INFO_STREAM(get_logger(), mode_size);
+    libcamera::StreamConfiguration &modecfg = cfg->at(1);
+    modecfg.size = mode_size;
+
+    modecfg.pixelFormat = libcamera::formats::SBGGR12_CSI2P;
```
larsll, line 376: This line is likely problematic, potentially not compatible with all cameras etc.

larsll: It might be an option to remove it entirely, and let libcamera find a suitable one. It seems like removing the line is ok.

src/CameraNode.cpp
```diff
@@ -357,6 +368,19 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options)
...
+
+    RCLCPP_INFO_STREAM(get_logger(), "Mode configuration: " << modecfg.toString());
+
+    set_parameter(rclcpp::Parameter("mode_width", int64_t(modecfg.size.width)));
+    set_parameter(rclcpp::Parameter("mode_height", int64_t(modecfg.size.height)));
```
christianrauch, line 381: Do you see a point in setting the parameters again? The node does not dynamically react to changes in the stream configuration. Therefore, all parameters related to the stream configuration (dimensions, format, role) are read-only and only read once at the beginning.

src/CameraNode.cpp
```diff
@@ -218,6 +218,8 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options)
   param_descr_ro.read_only = true;
   declare_parameter<int64_t>("width", {}, param_descr_ro);
   declare_parameter<int64_t>("height", {}, param_descr_ro);
+  declare_parameter<int64_t>("mode_width", {}, param_descr_ro);
+  declare_parameter<int64_t>("mode_height", {}, param_descr_ro);
```
christianrauch, line 222: What is the meaning of `mode` in the parameter prefix? I think these new parameters also need to be documented, as it is not clear what they do without looking at the code.

src/CameraNode.cpp
```diff
@@ -297,19 +299,27 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options)
...
+
+  if (!mode_size.isNull()) {
+    roles.push_back(libcamera::StreamRole::Raw);
+  }
+
   // configure camera stream
-  const libcamera::StreamRole role = get_role(get_parameter("role").as_string());
   std::unique_ptr<libcamera::CameraConfiguration> cfg =
-    camera->generateConfiguration({role});
+    camera->generateConfiguration({roles});
```
christianrauch, line 313: Since `roles` is already a list, wrapping them again in curled braces (`{}`) is unnecessary.

src/CameraNode.cpp
```diff
@@ -297,19 +299,27 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options)
...
     throw std::runtime_error("failed to acquire camera");
 
+  std::vector<libcamera::StreamRole> roles;
+
+  roles.push_back(get_role(get_parameter("role").as_string()));
+  const libcamera::Size mode_size(get_parameter("mode_width").as_int(), get_parameter("mode_height").as_int());
+
+  if (!mode_size.isNull()) {
+    roles.push_back(libcamera::StreamRole::Raw);
+  }
```
christianrauch, line 309: What happens if a user already selected `raw` for the `role`? Are we then adding the same role twice with potentially different conflicting configurations?

christianrauch (commented):
Thanks for the PR.

Can you add a before/after comparison to this PR? This will make it more clear what the new parameter do.

The new parameters and their behaviour and interaction with the other sizes and stream roles needs documentation, if this should make it into the upstream code. I think having two sets of sizes is confusing as it is not clear from the parameter names what the effect is.

src/CameraNode.cpp
christianrauch: Using this with another camera that does not support that format, such as a regular UVC/V4L2 camera, will throw a "failed to generate configuration" exception. So either the format needs to be exposed via another parameter or libcamera is selecting one.

christianrauch:
Have you had a look at the `std::optional<SensorConfiguration> sensorConfig;` inside the `CameraConfiguration`? This seems to store the sensor configuration, such as the binning, size and crop region. Can you achieve the same effect by setting those values?

src/CameraNode.cpp
larsll: The term mode seems to be the term for this -- and is a short for sensor mode; 
* https://www.raspberrypi.com/documentation/computers/camera_software.html#common-options
* https://github.com/raspberrypi/picamera2/discussions/756

We could opt for a style similar to the rpicam tools; where mode is `width:height:bit:pack`, and maybe calling it raw_mode or sensor_mode would make abit more sense.

src/CameraNode.cpp
larsll: I agree, a check should be made, that if role is raw then this is not having an effect (and then ignore it).

src/CameraNode.cpp
larsll: This came from @Woojin-Crive original code; I am not sure what it does - possibly it updates the parameter (allowing you to read out the value) if something changed. As it does not fit with parameter handling in general I agree we can delete it.

src/CameraNode.cpp
larsll: I tested, it should not be fixed like this. (And 'funnily' -- even setting it was overwritten because my CM4 + V2 camera went for `1640x1232-SBGGR10_CSI2P` (see log cut out above)

src/CameraNode.cpp
christianrauch: According to the first source, "sensor mode" refers to what the node uses as format, e.g. `SRGGB10_CSI2P`:
> Sensor mode identifiers have the following form: S<Bayer order><Bit-depth>_<Optional packing> : <Resolution list>

How are the [rpicam-apps](https://github.com/raspberrypi/rpicam-apps) terming these parameters?

src/CameraNode.cpp
christianrauch: Alternatively, we could remove `raw` from the options for `role`. The device specific raw formats aren't supported by the ROS message anyway, and I am not sure if there is a benefit of publishing in Bayer format and then doing the debayering somewhere else.

src/CameraNode.cpp
christianrauch: If it is supposed to update the parameter such that you can see the actual "mode size" with the ROS parameter API, then this does not work.

E.g. setting a sensor resolution of 1640x1232:
```
ros2 run camera_ros camera_node --ros-args -p height:=480 -p width:=640 -p mode_width:=1640 -p mode_height:=1232
```
will be adjusted to `2304x1296-BGGR_PISP_COMP1`, but
```
ros2 param dump /camera
```
still shows:
```
/camera:
  ros__parameters:
    mode_height: 1232
    mode_width: 1640
```

christianrauch:
When you update the PR, can you rebase and squash your commits? I merged a change that avoids referring to the read-only parameters multiple times. Have a look at this and try following the same with the new read-only parameters that you want to add.

src/CameraNode.cpp
larsll: For now I have edited the code to not use my changes if the first role is a Raw.

christianrauch:
> Have you had a look at the `std::optional<SensorConfiguration> sensorConfig;` inside the `CameraConfiguration`? This seems to store the sensor configuration, such as the binning, size and crop region. Can you achieve the same effect by setting those values?

Setting the `sensorConfig` (`SensorConfiguration`)
```C++
  libcamera::SensorConfiguration sensor;
  sensor.outputSize = libcamera::Size(4608, 2592);
  sensor.bitDepth = 10;

  if (sensor.isValid())
    cfg->sensorConfig = sensor;
  else
    throw std::runtime_error("invalid sensor configuration");
```
before the `CameraConfiguration` is validated, indeed has the same effect:
```
INFO Camera camera.cpp:1202 configuring streams: (0) 160x120-XRGB8888
INFO RPI pisp.cpp:1484 Sensor: /base/axi/pcie@120000/rp1/i2c@88000/imx708@1a - Selected sensor format: 4608x2592-SBGGR10_1X10 - Selected CFE format: 4608x2592-PC1B
```
without the need to manage an additional stream.

But the size and bit depth have to match exactly for the validation to pass.

larsll:
Back. I would then suggest to do a parameter `sensor`, which is width:height:bitdepth using what you wrote? I am not worried about the invalid sensor configuration, as it is something you set up once for a camera, in a given setting.

christianrauch:
> Back. I would then suggest to do a parameter `sensor`, which is width:height:bitdepth using what you wrote?

Yeah, fine with me. Can you link here where the format string is used that way? How do they incorporate the other parameters, such as `binning` and `skipping`, into the config?

An alternative would be to have the parameters `width`, `height`, `bitdepth` inside the parameter namespace `sensor`, i.e. such that you can set them via `sensor.width`, `sensor.height`, `sensor.bitdepth` on the command line. But that is quite verbose, since you have to set all three parameters at once if you want to change the sensor configuration.

> I am not worried about the invalid sensor configuration, as it is something you set up once for a camera, in a given setting.

You just have to know the valid values for each of the parameter, since there is no automatic adjustment. And it is not so obvious where to get this information from.

I think it would be nice to show the valid values on the screen, similar to what `cam -c1 -I --stream role=raw` shows, if configuration fails. But this would again require to at least generate and read the raw stream configuration once (`generateConfiguration`) without using the stream. Not sure if there is another way to get this information.

larsll:
Unfortunately SensorConfiguration is not an option for ROS Humble, as ros-humble-libcamera is version 0.1.0; see https://github.com/raspberrypi/libcamera/blob/v0.1.0/include/libcamera/camera.h, where SensorConfiguration is missing. (It seems to be included in some RPi specific patches, otherwise starting in v0.2.0.

larsll:
> I think it would be nice to show the valid values on the screen, similar to what `cam -c1 -I --stream role=raw` shows, if configuration fails. But this would again require to at least generate and read the raw stream configuration once (`generateConfiguration`) without using the stream. Not sure if there is another way to get this information.

Looking at picamera2 -- doing generateConfiguration is what they do:
https://github.com/raspberrypi/picamera2/blob/fb031cb49cb2af4dab7439be7a90784c9d3db1ab/picamera2/picamera2.py#L483

christianrauch:
> Unfortunately SensorConfiguration is not an option for ROS Humble, as ros-humble-libcamera is version 0.1.0; see https://github.com/raspberrypi/libcamera/blob/v0.1.0/include/libcamera/camera.h, where SensorConfiguration is missing. (It seems to be included in some RPi specific patches, otherwise starting in v0.2.0.

Either we make this an libcamrea >= 0.2 only feature or we just leave libcamea 0.1 and ROS humble support for bloom packages behind. libcamera has high requirements on the meson version, the build farm only allows using system packages for build tool, and the RHEL 8 support here is the lower bound of the meson support we have to deal with on the build farm.

Options:
1. bump the version requirement for libcamera to support `SensorConfiguration` and then advice ROS humble users to compile libcamea from source, using whatever meson version from PyPI required for libcamera
2. use preprocessor macros to disable that feature at compile time on libcamera <= 0.1 (see `LIBCAMERA_VER_GE`)

I am in favour of disabling that feature and its parameters at compile time via `LIBCAMERA_VER_GE`. This is already done a couple of times in the code to support multiple libcamera versions.

christianrauch:
Btw, if this changes the FoV independently of the output resoltion, that property has to be encoded in the calibration file name.

Currently, the calibration file name is a concatenation of the model (property `libcamera::properties::Model`), camera name (`libcamera::Camera::id()`), and the stream resolution (`libcamera::StreamConfiguration::size`), e.g.: `IntegratedCamera_IntegratedC___SB__PCI0_XHC__RHUB_HS08_8_1_0_5986_2115_1280x720.yaml`. Having the same stream resolution with different sensor modes will create conflicts as this will map to the same file name. Hence, we should append the sensor configuration or anything else that affects the FoV and image centre to the file name to avoid this conflict.

christianrauch:
> > I think it would be nice to show the valid values on the screen, similar to what `cam -c1 -I --stream role=raw` shows, if configuration fails. But this would again require to at least generate and read the raw stream configuration once (`generateConfiguration`) without using the stream. Not sure if there is another way to get this information.
> 
> Looking at picamera2 -- doing generateConfiguration is what they do: https://github.com/raspberrypi/picamera2/blob/fb031cb49cb2af4dab7439be7a90784c9d3db1ab/picamera2/picamera2.py#L483

Then let's do the same. When the selected `SensorConfiguration` is invalid or the configuration fails, show a warning with the supported sensor modes like for the other parameters.

larsll:
> Options:
> 
> 1. bump the version requirement for libcamera to support `SensorConfiguration` and then advice ROS humble users to compile libcamea from source, using whatever meson version from PyPI required for libcamera
> 2. use preprocessor macros to disable that feature at compile time on libcamera <= 0.1 (see `LIBCAMERA_VER_GE`)
> 
> I am in favour of disabling that feature and its parameters at compile time via `LIBCAMERA_VER_GE`. This is already done a couple of times in the code to support multiple libcamera versions.

A third option is obviously to have two different implementations; using SensorConfiguration for higher versions of LibCamera, and the RAW stream for lower versions. It will be a bit of 'ifdef' spaghetti though.

My personal need is to have support for Ubuntu 22.04 (meaning ROS Humble), Compute Module 4 and Camera Module v2; where for some reason legacy support is no longer working, so I would need something compatible with that (or else I would have to maintain a fork, which I don't want to do.

christianrauch:
> > Options:
> > 
> > 1. bump the version requirement for libcamera to support `SensorConfiguration` and then advice ROS humble users to compile libcamea from source, using whatever meson version from PyPI required for libcamera
> > 2. use preprocessor macros to disable that feature at compile time on libcamera <= 0.1 (see `LIBCAMERA_VER_GE`)
> > 
> > I am in favour of disabling that feature and its parameters at compile time via `LIBCAMERA_VER_GE`. This is already done a couple of times in the code to support multiple libcamera versions.
> 
> A third option is obviously to have two different implementations; using SensorConfiguration for higher versions of LibCamera, and the RAW stream for lower versions. It will be a bit of 'ifdef' spaghetti though.

I am not a fan of having two paths to the same function. That's just a additional maintainance burden. If a solution works for both ROS / libcamera versions, I would rather only maintain that one solution.

> My personal need is to have support for Ubuntu 22.04 (meaning ROS Humble), Compute Module 4 and Camera Module v2; where for some reason legacy support is no longer working, so I would need something compatible with that (or else I would have to maintain a fork, which I don't want to do.

I see. If you cannot use the newer version and are the one goint to implement this, then you are leaving me no choice but to accept your implementation :-)

With "legacy support" do you mean support for the legacy camera stack? This isn't supported on any platform with newer Raspberry Pi OS any more. May I ask why you cannot use Ubuntu 24.04 on the Raspberry Pi Compute Module 4? Is compiling from source also not an option for you?

larsll:
> With "legacy support" do you mean support for the legacy camera stack? This isn't supported on any platform with newer Raspberry Pi OS any more. 

Yes. For some obscure reason the RPi4 / Ubuntu 22.04 combo supports Legacy Camera, so I can activate that and grab the pictures using a [basic camera package provided by AWS](https://github.com/aws-deepracer/aws-deepracer-camera-pkg).

As we want to reduce the footprint of [the solution](https://github.com/aws-deepracer-community/deepracer-custom-car) we also want to support the CM4 with a custom carrier board, and for some strange reason Legacy Camera is not working with the CM4. I thought I had a solution when I found camera_ros, only to see that libcamera ends up cropping the picture... 

> On second thought, if you are using the Camera Module v2 on Ubuntu, I assume you are already compiling libcamera from source. Or does the binary bloomed libcamera package in the ROS repo support the Camera Module v2? If you already compile libcamera from source, doing the same for the node should not be a problem.

`ros-humble-libcamera` works with Ubuntu 22.04/CM4/Camera Module v2 as-is, no need to compile it. The preferred solution is to use the package for simplicity / reduce future work. As the project patches and compiles up a full DeepRacer package-set and bundles it as a deb-package, we can easily compile a branch/fork of camera_ros (but comes with the future maintenance penalty). 

> May I ask why you cannot use Ubuntu 24.04 on the Raspberry Pi Compute Module 4? Is compiling from source also not an option for you?

We are looking to support that, as support for RPi5, CM5 and other cameras is desirable as well, but we have quite a bit of OS-level integration (network configuration, USB-OTG etc.) that will require work. 

> I see. If you cannot use the newer version and are the one goint to implement this, then you are leaving me no choice but to accept your implementation :-)
 
I suggest to do the following:
- Define the input parameter as `sensor`; initial values are `width:height`, but with future expansion to `width:height[:somethingelse]` being backward compatible. This will enable a change of implementation without breaking it for the user.
- At some future point have a separate ros-humble branch with the main branch supporting ros-jazzy and whatever comes next. This is in line with how other packages maintain backwards compatibility with older ROS versions.

larsll:
> Btw, if this changes the FoV independently of the output resoltion, that property has to be encoded in the calibration file name.
> 
> Currently, the calibration file name is a concatenation of the model (property `libcamera::properties::Model`), camera name (`libcamera::Camera::id()`), and the stream resolution (`libcamera::StreamConfiguration::size`), e.g.: `IntegratedCamera_IntegratedC___SB__PCI0_XHC__RHUB_HS08_8_1_0_5986_2115_1280x720.yaml`. Having the same stream resolution with different sensor modes will create conflicts as this will map to the same file name. Hence, we should append the sensor configuration or anything else that affects the FoV and image centre to the file name to avoid this conflict.

Ok. I will look at how `cam` is doing this and replicate it if suitable.

larsll:
More dependency complexity; 
* Raspberry Pi 4 + Pi Camera v2 + Ubuntu 22.04 + ros-humble-libcamera works.
* Raspberry Pi 5 + Pi Camera v2 + Ubuntu 24.04 + ros-jazzy-libcamera does NOT work.

Getting the latter combination to work indeed requires custom build of libcamera, off the Raspberry Pi fork. 

Could you point me to where the building of the ros-*-libcamera happens? Some adjustments would probably be required there to make things work out of the box.

christianrauch:
> More dependency complexity;
> 
> * Raspberry Pi 4 + Pi Camera v2 + Ubuntu 22.04 + ros-humble-libcamera works.
> * Raspberry Pi 5 + Pi Camera v2 + Ubuntu 24.04 + ros-jazzy-libcamera does NOT work.
> 
> Getting the latter combination to work indeed requires custom build of libcamera, off the Raspberry Pi fork.

You are probably aware of the different "libcamera variants". The package in the ROS repo is built from the official upstream version at https://git.libcamera.org/libcamera/libcamera.git. But the development of the Raspberry Pi support happens in the @raspberrypi fork at https://github.com/raspberrypi/libcamera. That support is supposed to be merged into the upstream at some point, but this process seems to take a long time. See for example the patch series ["Raspberry Pi: Add support for Pi 5 (PiSP)"](https://patchwork.libcamera.org/project/libcamera/list/?series=4944)  by @naushir, which is still under review. That means that the upstream repo, and hence the bloomed package, usually only supports older Raspberry Pi models and camera modules.

> Could you point me to where the building of the ros-*-libcamera happens? Some adjustments would probably be required there to make things work out of the box.

That package is "bloomed" using https://github.com/ros-infrastructure/bloom/pull/691 directly on the upstream repo. This generates a Debian package specification, stores it in the release repo at https://github.com/ros2-gbp/libcamera-release, and then the ROS build farm is generating the binary package from this specification. You can also use bloom locally to generate and build the Debian package: https://docs.ros.org/en/jazzy/How-To-Guides/Building-a-Custom-Deb-Package.html.

But if you want Raspberry Pi 5 support out-of-the-box in the upstream repo and hence the ROS repo, you should nag @naushir to get the Raspberry Pi 5 patches merged upstream :-)

larsll:
> But if you want Raspberry Pi 5 support out-of-the-box in the upstream repo and hence the ROS repo, you should nag @naushir to get the Raspberry Pi 5 patches merged upstream :-)
Thanks for the references. I will have a look. Once I get serious about doing Jazzy + Pi 5 support I might just build a drop-in replacement package for ros-jazzy-libcamera and put it in our APT repository with our other patched community work for DeepRacer.

README.md
```diff
@@ -123,20 +123,32 @@ The camera stream is configured once when the node starts via the following stat
 | `role`            | `string`              | configures the camera with a `StreamRole` (possible choices: `raw`, `still`, `video`, `viewfinder`) [default: `viewfinder`] |
 | `format`          | `string`              | a `PixelFormat` that is supported by the camera [default: auto]                                                             |
 | `width`, `height` | `integer`             | desired image resolution [default: auto]                                                                                    |
+| `sensor`          | `string` in format `width:height`   | desired raw sensor format resolution [default: auto]                                                            |
```
christianrauch, line 126: Minor: Move the expected format to the description column. See the `role` parameter as example.

README.md
```diff
@@ -123,20 +123,32 @@ The camera stream is configured once when the node starts via the following stat
...
 
 The node avoids memory copies of the image data by directly mapping from a camera pixel format to a ROS image format, with the exception of converting between "raw" and "compressed" image formats when requested by the user. As an effect, not all pixel formats that are supported by the camera may be supported by the ROS image message. Hence, the options for `format` are limited to pixel formats that are supported by the camera and the raw ROS image message.
 
+#### Sensor Modes and Cropping
+
+Some camera modules (e.g. Raspberry Pi Camera Modules) provides different sensor modes, some of which have a limited field-of-view, e.g. by allowing for the selection of a 3:2 image from a 16:9 sensor, a more native 'zoom' function etc. Unless we specify the sensor mode we want to use libcamera will automatically select one, and it is not guaranteed that this will be a full sensor mode, leading to cropping (which provides a zooming effect) of the picture.
```
christianrauch, line 144: "provides" -> "provide"

README.md
```diff
@@ -123,20 +123,32 @@ The camera stream is configured once when the node starts via the following stat
...
+
+As an example: Picking the commonly used 640x480 resolution with the Raspberry Pi Camera Module v2 gives a zooming effect of ~2.5x, greatly reducing the FoV. With the v1 Module there is no such zooming effect as the v1 camera only offers full FoV modes in the 4:3 aspect ratio.
+
+With the Camera Module v2 it is suggested to use a sensor resolution of 1640x1232 when one needs lower resolution 4:3 images (e.g. 640x480, 480x320 etc.), and 1640x922 for lower resolution 16:9 images.
```
christianrauch, line 148: Thanks for such a detailed description. But IMHO, this is a bit too Camera Module v1 and v2 specific. I think it is not required to go in such detail and it would be sufficient to explain the general relation of the discrete sensor modes and the variable output resolutions, and the problem with the automatic selection of the sensor mode and its effect on the FoV. Link external resources as much as you can.

README.md
```diff
@@ -123,20 +123,32 @@ The camera stream is configured once when the node starts via the following stat
...
+
+See the [PiCamera Documentation](https://picamera.readthedocs.io/en/release-1.13/fov.html#sensor-modes) for a visualization of the issue, and [section 4.2.2.3 in the PiCamera2 Documentation](https://datasheets.raspberrypi.com/camera/picamera2-manual.pdf) for further explanations.
```
christianrauch, line 150: I guess the information is still valid, but this links to the "old" MMAL-based `picamera` camera stack and not the new libcamera-based `picamera2`. Is there similar informatoin about the effect of the sensor mode on the FoV available for `picamera2`?

christianrauch (commented):
Thanks for your work. I am OK with the general idea and its implementation via the raw stream configuration. But there are still some ways to optimise the code to make this better maintainable in the future.

README.md
larsll: The PDF for picamera2 (section 4.2.2.2) is quite good:

>  Image sensors normally have a number of different modes in which they can operate. Modes are distinguished by
 producing different output resolutions, some of them may give you a different field of view, and there’s often a trade-off
 where lower resolution sensor modes can produce higher framerates. In many applications, it’s important to understand
 the available sensor modes, and to be able to choose the one that is the most appropriate.

larsll:
Updated as per request - left it for you to resolve the comments

src/CameraNode.cpp
```diff
@@ -299,18 +339,26 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options)
   if (camera->acquire())
     throw std::runtime_error("failed to acquire camera");
 
+  std::vector<libcamera::StreamRole> roles;
+  roles.push_back(role);
+
+  // Add the RAW role if the sensor_size is defined
+  if (!sensor_size.isNull() && roles.front() != libcamera::StreamRole::Raw) {
```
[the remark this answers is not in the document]
christianrauch: Why do you still want to use `roles.front()` instead of the explicit `role` here?

src/CameraNode.cpp
```diff
@@ -357,6 +405,12 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options)
     scfg.size = size;
   }
 
+  if (!sensor_size.isNull() && roles.front() != libcamera::StreamRole::Raw) {
```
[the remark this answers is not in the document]
christianrauch: Why do you still want to use `roles.front()` instead of the explicit `role` here?

christianrauch (changes requested):
Almost there. I just left comments about the wording in the documentation and the use of `roles.front()`. Also, can you squash and rebase your commits? I would like to keep the git history clean of "fixup" commits.

src/CameraNode.cpp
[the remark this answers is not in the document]
larsll: Removed now.

src/CameraNode.cpp
```diff
@@ -299,18 +339,26 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options)
...
   // configure camera stream
   std::unique_ptr<libcamera::CameraConfiguration> cfg =
-    camera->generateConfiguration({role});
+    camera->generateConfiguration(roles);
 
   if (!cfg)
     throw std::runtime_error("failed to generate configuration");
 
-  assert(cfg->size() == 1);
+  assert(cfg->size() >= 1);
```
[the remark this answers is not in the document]
larsll: Last open point (I think) - I suspect it is obsolete, and more a final check to avoid doing cfg->at(0) if it is empty. Similar topic at line 402, where checking `!sensor_size.isNull()` implicitly checks if cfg->at(1) will work in L403.

larsll:
All comments built in. Rebased (I normally do it via the built-in GitHub functionality when merging, this is why I did not do it).

Final open  question is https://github.com/christianrauch/camera_ros/pull/100#discussion_r1992349508

src/CameraNode.cpp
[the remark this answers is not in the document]
christianrauch: > In the end it is a question about preference/style - do you want to rely on 'knowing' the state, or add checks to be twice sure?

Yes, it's mostly about knowing the state and preventing crashing the process or reading garbage from memory. libcamera does not throw exceptions as far as I know, so accessing a list like this beyond its size either reads random data or crashes the process. But by calling `generateConfiguration` with a known number of roles, I am not sure if it can ever happen that the length of the returned stream configurations can be different than the given number of given roles. If we know how many roles we provide, we also know how many configurations we get, I guess. But in `Release` mode, those `assert`s will be compiled away anyway.

So anyway, you could also just move a check for `cfg->size() != 0` into the check for `!cfg` above and throw the same exception.

> Considering this led to adjusting check in L443 to `!sensor_size.isNull() && role != libcamera::StreamRole::Raw` as well as this is the condition for using the sensor_size.

src/CameraNode.cpp
```diff
@@ -221,6 +249,13 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options)
   const uint32_t h = declare_parameter<int64_t>("height", {}, param_descr_ro);
   const libcamera::Size size {w, h};
 
+  // Raw format dimensions
+  rcl_interfaces::msg::ParameterDescriptor param_descr_sensor_mode;
+  param_descr_sensor_mode.description = "raw mode of the sensor";
+  param_descr_sensor_mode.additional_constraints = "string in format [width]:[height]";
+  param_descr_sensor_mode.read_only = true;
+  const libcamera::Size sensor_size = get_sensor_format(declare_parameter<std::string>("sensor_mode", {}, param_descr_sensor_mode));
```
christianrauch, line 257: All your `runtime_error` are not handled here. This will make the node quit when an invalid format is provided. Is this intended?

src/CameraNode.cpp
```diff
@@ -150,6 +151,33 @@ get_role(const std::string &role)
...
+    catch (const std::invalid_argument &) {
+      throw std::runtime_error("Invalid size format! Expected [width]:[height] but got " + format_str);
+    }
+    catch (const std::out_of_range &) {
+      throw std::runtime_error("Invalid size format! Expected [width]:[height] but got " + format_str);
+    }
+  }
+  else {
+    throw std::runtime_error("Invalid size format! Expected [width]:[height] but got " + format_str);
+  }
```
christianrauch, line 179: You are throwing the same exception three times. This could be optimised, i.e. figure out if the format is valid and if not, throw an exception or warning.

christianrauch (changes requested):

src/CameraNode.cpp
larsll: Yes - similar to the `get_role` function. If you specifically ask for something that is not possible I would rather that the program quits than it proceeds 'silently' with different values.

src/CameraNode.cpp
larsll: What is preferred here? We can force the number of digits via the regex which will prevent out_of_range and invalid_argument -- do you still want the try/catch?

This should work equally well, without the repeated throws.
```
libcamera::Size
get_sensor_format(const std::string &format_str)
{
  if (format_str.empty()) {
    return {};
  }

  const std::regex pattern(R"((\d{1,11}):(\d{1,11}))");
  std::smatch match;

  if (std::regex_match(format_str, match, pattern)) {
    const unsigned long width = std::stoul(match[1].str());
    const unsigned long height = std::stoul(match[2].str());

    // Check for overflow
    if (width > std::numeric_limits<uint32_t>::max() || height > std::numeric_limits<uint32_t>::max()) {
      throw std::runtime_error("Width or height exceeds maximum value of " + std::to_string(std::numeric_limits<uint32_t>::max()));
    }

    return libcamera::Size {static_cast<uint32_t>(width), static_cast<uint32_t>(height)};
  }
  else {
    throw std::runtime_error("Invalid size format! Expected [width]:[height] but got " + format_str);
  }
}
```

src/CameraNode.cpp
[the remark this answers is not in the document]
larsll: This was changed into a single evaluation.

src/CameraNode.cpp
christianrauch: Still try/catch all the exceptions. I just meant that `throw std::runtime_error("Invalid size format! Expected [width]:[height] but got " + format_str);` three times is unnecessary code repetition. I would just set a boolean flag if an error occurs and then check that flag at the end to throw this exception.

src/CameraNode.cpp
```diff
@@ -150,6 +151,31 @@ get_role(const std::string &role)
...
+  if (format_str.empty()) {
+    return {};
+  }
+
+  const std::regex pattern(R"((\d{1,19}):(\d{1,19}))");
+  std::smatch match;
+
+  if (std::regex_match(format_str, match, pattern)) {
+    const unsigned long width = std::stoul(match[1].str());
+    const unsigned long height = std::stoul(match[2].str());
```
christianrauch, line 166: [`std::stoul` may throw exceptions.](https://en.cppreference.com/w/cpp/string/basic_string/stoul) Previously you handled them here and then threw your own exceptions. Do you now want to have the caller of `get_sensor_format` handle them? If you use your own exceptions, it might still be useful to forward the information from `std::invalid_argument` and `std::out_of_range`.

src/CameraNode.cpp
```diff
@@ -150,6 +151,31 @@ get_role(const std::string &role)
...
+
+    // Check for overflow
+    if (width > std::numeric_limits<uint32_t>::max() || height > std::numeric_limits<uint32_t>::max()) {
+      throw std::runtime_error("Width or height exceeds maximum value of " + std::to_string(std::numeric_limits<uint32_t>::max()));
+    }
```
christianrauch, line 171: What is this check for? The `std::out_of_range` exceptions should be thrown before you can do this check.

christianrauch (changes requested):
In `get_sensor_format`, we have to either handle all the exceptions and give meaningful messages to the users, or not handle them and let the process quit with the default message of that unhandled exception.

src/CameraNode.cpp
larsll: Ok. As the positive flow returns directly we don't even need to set a boolean flag; if you arrive at the bottom something did not fit.
```
libcamera::Size
get_sensor_format(const std::string &format_str)
{
  if (format_str.empty()) {
    return {};
  }

  const std::regex pattern(R"((\d{1,19}):(\d{1,19}))");
  std::smatch match;

  if (std::regex_match(format_str, match, pattern)) {
    try {
      const unsigned long width = std::stoul(match[1].str());
      const unsigned long height = std::stoul(match[2].str());

      // Check for overflow
      if (width > std::numeric_limits<uint32_t>::max() || height > std::numeric_limits<uint32_t>::max()) {
        throw std::runtime_error("Width or height exceeds maximum value of " + std::to_string(std::numeric_limits<uint32_t>::max()));
      }

      return libcamera::Size {static_cast<uint32_t>(width), static_cast<uint32_t>(height)};
    }
    catch (const std::invalid_argument &) {
      // Do nothing - throw exception below
    }
    catch (const std::out_of_range &) {
      // Do nothing - throw exception below
    }
  }

  throw std::runtime_error("Invalid size format! Expected [width]:[height] but got " + format_str);

}
```

src/CameraNode.cpp
larsll: See the discussion above. Will re-add.

src/CameraNode.cpp
larsll: `stoul` returns `unsigned long`, whereas libcamera::Size wants `unsigned int`, hence out_of_range will trigger at a much higher number than we can accept. Given that we are already using a regexp, negative numbers are not possible, so we can use `stoi` instead. We will be limited to the positive half of `int`, and then cast it to `unsigned int`. The reduced max number should not have any practical meaning.

larsll: Updated to use `stoi`.

christianrauch: You are right. I assumed there would be a matching conversion function for unsigned integers. Weirdly enough, there is a `stoi` and `stol` for signed `int` and `long`, but only `stoul` for `unsigned long` and no `stoui` for `unsigned int`. See: https://en.cppreference.com/w/cpp/header/string.

Depending on whether you handle `std::out_of_range` in the function or not, it would be nice to throw the same exception to "simulate" the potential behaviour of `stoui`.

src/CameraNode.cpp
```diff
@@ -150,6 +151,34 @@ get_role(const std::string &role)
...
+  if (std::regex_match(format_str, match, pattern)) {
+    try {
+      const int width = std::stoi(match[1].str());
+      const int height = std::stoi(match[2].str());
+
+      return libcamera::Size {static_cast<unsigned int>(width), static_cast<unsigned int>(height)};
+    }
+    catch (const std::out_of_range &) {
+      throw std::runtime_error("Width or height exceeds maximum value of " + std::to_string(std::numeric_limits<int>::max()));
+    }
```
christianrauch, line 173: What does the "explanatory string" via `what()` contain here? Can't you just forward that information?

larsll: `e.what()` returns only `stoi`. Not very userfriendly. Have adjusted the error messages to clearly state it is `sensor_mode` that is wrong.

src/CameraNode.cpp
larsll: Will handle within to convert into user-friendly error message.

christianrauch (approved):

christianrauch:
Looks good now. Thanks a lot.

PR #120, opened 2025-05-08
[kilted] Update deprecated call to ament_target_dependencies

DLu:
As of the ROS 2 Kilted release [`ament_target_dependencies` is deprecated.](https://docs.ros.org/en/kilted/Releases/Release-Kilted-Kaiju.html#ament-target-dependencies-is-deprecated)

This PR updates the syntax. Caution should be used to ensure that the target branch is not used for other ROS 2 distros.

Note: This PR was generated by a bot script, but using the simple pattern matching of the [`ros_glint`](https://github.com/MetroRobots/ros_glint) library. No LLMs were used.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,8 +30,7 @@ add_library(utils OBJECT
   src/pretty_print.cpp
 )
 target_include_directories(utils PUBLIC ${libcamera_INCLUDE_DIRS})
-target_link_libraries(utils ${libcamera_LINK_LIBRARIES})
-target_link_libraries(utils
+target_link_libraries(utils ${libcamera_LINK_LIBRARIES}
   PUBLIC
   "sensor_msgs"
 )
@@ -48,8 +47,7 @@ add_library(param OBJECT
     src/ParameterConflictHandler.cpp
 )
 target_include_directories(param PUBLIC ${libcamera_INCLUDE_DIRS})
-target_link_libraries(param ${libcamera_LINK_LIBRARIES})
-target_link_libraries(param
+target_link_libraries(param ${libcamera_LINK_LIBRARIES}
   PUBLIC
   "rclcpp"
 )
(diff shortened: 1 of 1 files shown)


christianrauch:
If this breaks older ROS 2 distributions, couldn't you add an ament version check and then switch the behaviour based on this?

Also, can you please fix the compilation issue (see CI) and squash this together?

CMakeLists.txt
```diff
@@ -30,10 +30,10 @@ add_library(utils OBJECT
...
 )
 target_include_directories(utils PUBLIC ${libcamera_INCLUDE_DIRS})
-target_link_libraries(utils ${libcamera_LINK_LIBRARIES})
-ament_target_dependencies(
-  utils
-  "sensor_msgs"
+target_link_libraries(utils PUBLIC
+  ${libcamera_LINK_LIBRARIES}
+  ${sensor_msgs_TARGETS}
+  sensor_msgs::sensor_msgs_library
```
christianrauch, line 36: Why is the target `sensor_msgs::sensor_msgs_library` not part of `${sensor_msgs_TARGETS}`?

CMakeLists.txt
```diff
@@ -65,9 +64,8 @@ set(PACKAGE_DEPENDENCIES
   "camera_info_manager"
   "cv_bridge"
 )
-ament_target_dependencies(camera_component PUBLIC ${PACKAGE_DEPENDENCIES})
 target_include_directories(camera_component PUBLIC ${libcamera_INCLUDE_DIRS})
-target_link_libraries(camera_component PUBLIC ${libcamera_LINK_LIBRARIES})
+target_link_libraries(camera_component PUBLIC ${libcamera_LINK_LIBRARIES} ${sensor_msgs_TARGETS} camera_info_manager::camera_info_manager cv_bridge::cv_bridge rclcpp::rclcpp rclcpp_components::component rclcpp_components::component_manager sensor_msgs::sensor_msgs_library)
```
christianrauch, line 68: If this is already using `${sensor_msgs_TARGETS}`, why not continue with this pattern for the remaining targets, e.g. `${camera_info_manager_TARGETS}`, `${rclcpp_TARGETS}`? IMHO, this list is quite verbose compared to the previous solution. Also, I would prefer to keep this as a list with one element per line, as with the previous `PACKAGE_DEPENDENCIES` list.

We now have to maintain two lists, `PACKAGE_DEPENDENCIES` for `ament_export_dependencies` and the list of targets for `target_link_libraries`, and keep them in sync. Since the targets can be derived from the package name, there should be a macro doing this automatically. Can we please also use the same targets for `ament_export_dependencies`?

christianrauch (changes requested):
I like the general idea of using CMake commands instead of custom ament macros. But I still see an issue with the duplication of dependencies and their targets. Also, the CI is throwing errors again. Could you maybe test these PRs first, by proposing PRs against your own `main` branch to let the CI run? At least, this should be a `Draft` before you can actually request reviews.

christianrauch:
@DLu Are you still interested in this PR? If so, can you please address the issues and fix the CI?

christianrauch:
Superseded by https://github.com/christianrauch/camera_ros/pull/124.

PR #121, opened 2025-05-08
add documentatin on how to generate core dumps

christianrauch:
Additional information on how to generate core dumps as alternative to backtraces.


christianrauch:
@dbacher42 Since you were the first to generate a core dump with parts of these instructions, could you have a look at this PR and see if this makes sense to you?

PR #143, opened 2025-11-28
Use modern image_transport API

eigendude:
## Description

This PR updates camera_ros to use the image_transport API, currently tested with ROS 2 Kilted.

The image_transport package takes care of image compression, allowing this PR to be net-negative in lines of code. It also offers an expanded range of alternative transports, such as zstd compression and ffmpeg encoding.

New in the Kilted API is combined "camera" publishers/subscribers, which allows both image and camera info to be processed at the same time in the same function. The new camera API is used in this PR.

## Motivation and contexet

I recently had the need to test custom QoS profiles. I noticed that v4l2_camera recently exposed a way to control the QoS of publishers. It would be valuable if camera_ros could do the same.

v4l2_camera uses the new camera API, so I updated camera_ros likewise, resulting in the diff here. However, I finished my testing before fully exposing the publisher's QoS. Still, this change allows v4l2_camera-style QoS configuration in the future.

## How has this been tested?

I'm currently running this patch on my robotic Millennium Falcon, along with image rectification. The new transport topics successfully show up:

```
# ros2 topic list
/oasis/falcon/camera_info
/oasis/falcon/image_raw
/oasis/falcon/image_raw/compressed
/oasis/falcon/image_raw/compressedDepth
/oasis/falcon/image_raw/ffmpeg
/oasis/falcon/image_raw/theora
/oasis/falcon/image_raw/zstd
/oasis/falcon/image_rect
/oasis/falcon/image_rect/compressed
/oasis/falcon/image_rect/compressedDepth
/oasis/falcon/image_rect/ffmpeg
/oasis/falcon/image_rect/theora
/oasis/falcon/image_rect/zstd
```

Downstream, I'm using ORB_SLAM3 on the `image_raw` topic and apriltag_ros on the `image_rect` topic for robot pose estimation, control and visualization.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,7 @@ set(PACKAGE_DEPENDENCIES
   "sensor_msgs"
   "camera_info_manager"
   "cv_bridge"
+  "image_transport"
 )
 foreach(PKGDEP IN LISTS PACKAGE_DEPENDENCIES)
   find_package(${PKGDEP} REQUIRED)
@@ -79,6 +80,7 @@ target_link_libraries(camera_component PUBLIC
   ${sensor_msgs_TARGETS}
   camera_info_manager::camera_info_manager
   cv_bridge::cv_bridge
+  image_transport::image_transport
 )
 target_include_directories(camera_component PUBLIC ${libcamera_INCLUDE_DIRS})
 target_link_libraries(camera_component PUBLIC ${libcamera_LINK_LIBRARIES})
--- a/package.xml
+++ b/package.xml
@@ -15,6 +15,7 @@
   <depend>sensor_msgs</depend>
   <depend>camera_info_manager</depend>
   <depend>cv_bridge</depend>
+  <depend>image_transport</depend>
 
   <exec_depend>ros2launch</exec_depend>
   <exec_depend>ament_index_python</exec_depend>
--- a/src/CameraNode.cpp
+++ b/src/CameraNode.cpp
@@ -15,6 +15,7 @@
 #include <cv_bridge/cv_bridge.h>
 #endif
 #include <atomic>
+#include <image_transport/image_transport.hpp>
 #include <iostream>
 #include <libcamera/base/shared_fd.h>
 #include <libcamera/base/signal.h>
@@ -46,6 +47,7 @@
 #include <rclcpp/parameter.hpp>
 #include <rclcpp/parameter_value.hpp>
 #include <rclcpp/publisher.hpp>
+#include <rclcpp/qos.hpp>
 #include <rclcpp/time.hpp>
 #include <rclcpp_components/register_node_macro.hpp>
 #include <regex>
@@ -97,38 +99,18 @@ private:
 
   bool use_node_time;
 
-  rclcpp::Publisher<sensor_msgs::msg::Image>::SharedPtr pub_image;
-  rclcpp::Publisher<sensor_msgs::msg::CompressedImage>::SharedPtr pub_image_compressed;
-  rclcpp::Publisher<sensor_msgs::msg::CameraInfo>::SharedPtr pub_ci;
+  image_transport::CameraPublisher pub_camera;
 
   camera_info_manager::CameraInfoManager cim;
 
   ParameterHandler parameter_handler;
   std::string frame_id;
 
-#ifdef RCLCPP_HAS_PARAM_EXT_CB
-  // use new "post_set" callback to apply parameters
-  PostSetParametersCallbackHandle::SharedPtr param_cb_change;
-#else
-  OnSetParametersCallbackHandle::SharedPtr param_cb_change;
-#endif
-
-  // compression quality parameter
-  std::atomic_uint8_t jpeg_quality;
-
   void
   requestComplete(libcamera::Request *const request);
 
   void
   process(libcamera::Request *const request);
-
-  void
-  postParameterChange(const std::vector<rclcpp::Parameter> &parameters);
-
-#ifndef RCLCPP_HAS_PARAM_EXT_CB
-  rcl_interfaces::msg::SetParametersResult
-  onParameterChange(const std::vector<rclcpp::Parameter> &parameters);
-#endif
 };
 
 RCLCPP_COMPONENTS_REGISTER_NODE(camera::CameraNode)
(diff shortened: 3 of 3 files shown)


christianrauch (changes requested):
What is the motivation for this PR? Is this only about reducing the code? There is a reason why the compressed and raw data is handled this way. We want to prevent unessary image conversions. If this is just about QoS settings, you should be able to adjust the settings for the topics.

This also has to support older ROS 2 distros and cannot just use API from ROS 2 kilted onwards.

christianrauch:
> > There is absolutely no point on decompressing and recompressing the data. It wasts resources and reduces the image quality.
> 
> I guess that was my point. Capturing jpeg directly leads to impacted image quality regardless of how it's consumed. If the resource savings aren't worth it, then images shouldn't be captured as jpeg.

But this impact in quality is up to the user to decide. And it is also a question of the bandwidth. If you capture a raw data stream on a low-cost webcam, you may see rolling shutter effects due to the delay in transporting the raw uncompressed stream. This can be avoided by selecting a compressed data stream. Some cameras provide this compressed datastream, and who are we to tell users not to use this compressed datastream coming directly from the camera?

> But this point is only valid for today's PIs with high CPU speeds. The point is invalidated if the resource savings _are_ worth it, such as for low powered devices.

It is never worth to re-compress a JPEG image. Even if you have amble resources, decompressing the compressed datastream from the camera to a raw buffer and then compressing it back to a JPEG image will still reduce the quality.

> In that case, there's two paths forward: Keep the jpeg codepath and publish in parallel to e.g. `image_raw/libcamera_jpeg`, or modify image_transport to accept pre-compressed images for efficiency. With the first option, we lose one advantage of the camera API (publishing both messages at once). With the second option, we've expanded the scope a bunch by working with an upstream project.

The `CameraPublisher` only hides the fact that the raw and compressed images are published on separate topics behind a convinient API. The single API call makes it look like they are published at once, but there is not ROS API to publish on multiple topics at once. It is still publishing messages on topics using individual publishers.

If it would be possible to interface with the `CameraPublisher` to make it reuse our raw and compressed image buffers, it would safe us a lot of custom code. But this is not possible as far as I can see and this is the sole reason why the raw and compressed image data is directly reused for the ROS image messages without additional conversion or other preprocessing.

> > This also has to support older ROS 2 distros and cannot just use API from ROS 2 kilted onwards.
> 
> This is another downside of the change, it only builds against Kilted onward. This also has a solution, used by a large part of the ROS ecosystem, and that's branching. So not unsolvable but maybe not ideal.

But actually, the CI passes for older distros. So I am not sure if there is not a backward compatibility function or similar.

> I think we've covered enough downsides that the benefits of using the image_transport package aren't really worth it. The only argument left is for modernization, as someday someone might want better quality with zstd images or an h264 stream over a limited WiFi connection, or a future unseen use case, and you don't want to have to write code for that.

If this is just about compressing the data in another way, then you can also just use any composable node that compresses or decompresses the message via intraprocess communication within the same node manager. This provides the most flexibility and will allow you to use GPU-accelerated de-/compression that is not part of the `image_transport` package.

Example composable nodes:

- [`image_transport::Republisher`](https://docs.ros.org/en/ros2_packages/jazzy/api/image_transport/generated/classimage__transport_1_1Republisher.html)
- [`nvidia::isaac_ros::h264_decoder::EncoderNode`](https://github.com/NVIDIA-ISAAC-ROS/isaac_ros_compression/blob/main/isaac_ros_h264_encoder/include/isaac_ros_h264_encoder/encoder_node.hpp) and [`nvidia::isaac_ros::h264_decoder::DecoderNode`](https://github.com/NVIDIA-ISAAC-ROS/isaac_ros_compression/blob/main/isaac_ros_h264_decoder/include/isaac_ros_h264_decoder/decoder_node.hpp)

Coming back to your PR: What user-facing features are you exactly missing? Is this just about the QoS configuration? If so, can you just replicate the parameters for the topics that we already have?

eigendude:
> you may see rolling shutter effects due to the delay in transporting the raw uncompressed stream

Good point, I hadn't considered that

> but there is not ROS API to publish on multiple topics at once. It is still publishing messages on topics using individual publishers.

Right, the image_transport camera API is just a message_filter wrapper around the two topics. This is very valuable for subscribers, because they don't need their own message_filter, or hacky alternative.

But for publishers? There's no difference under the hood, and the only reason would be small code elegance.

> If it would be possible to interface with the CameraPublisher to make it reuse our raw and compressed image buffers, it would safe us a lot of custom code. But this is not possible as far as I can see and this is the sole reason why the raw and compressed image data is directly reused for the ROS image messages without additional conversion or other preprocessing.

Right, this feature would have to be added to image_transport for us to use it.

> composable node that compresses or decompresses the message via intraprocess communication within the same node manager

Good point, this solves the problem, and with zero-copy.

> image_transport::Republisher

And my solution for zstd and ffmpeg transports already exists! Thanks for pointing this out.

---

Conclusion: the code doesn't change unless image_transport gets such a feature.

Moreover, we would be worse off with this PR. If the jpeg stream were mapped to another topic, it wouldn't be seen by image_transport. My motivation for noticing this: I need to reduce my WiFi bandwidth to handle "Netflix Hour". Instead of configuring image_transport, I'll use the jpeg feature and crank the quality way down.

The RPi 5 running the camera node at 1080p, 15fps stabilizes around 12% CPU. I'm interested to see how low libcamera jpeg can take this, especially because increasing CPU reduces my battery's runtime.

eigendude:
> What is a "Netflix Hour"?

Lol, this is my idiom for that time in the late afternoon when everyone is home from work and streaming. There's only one air; Wi-Fi is a shared medium and only one device per group of frequency channels can send data at a time. So when everyone streams in a residential setting at the same time, WiFi slows down for everyone, including my robot's HD camera stream. I use "Netflix Hour" to refer to busiest time for WiFi where there's the most competition for traffic, where you see lots of fragged, dropped and retransmitted packets.

If you can solve for netflix hour, you can handle the most adversarial WiFi environment you can imagine 😉 

> In the end, all the compressing functions use OpenCV's `cv::imencode`

I discovered this yesterday while reading the code. I saw `MJPEG` format and mistakenly assumed the RPi camera could compress. But no, it's opencv turtles all the way down.

> just with a compression quality parameter

I'm sure image transport plugins have parameters too, but quick searching didn't uncover them, so I'm happy the param is exposed here. I'm glad to make use of it now.

I'm currently setting the quality to 60. It cut bandwidth from 10MB/s to 2.5MB/s. And I still have room to downscale from 1080p if I lose my lust for Full HD. With publishing queue size of 5 instead of 1, about half the frames get through netflix hour wifi, with only occasional half-second freezes. Fine for my nav algos.

Speaking of queue size, I did a lot of optimizing for bad wifi applications. I'll share the research, my results, and ideally a feature request to expose a `queue_size` parameter for publisher history depth.

> If you want to efficiently compress the images, you probably have to use the Raspberry Pis GPU or some hardware encoders to encode the image data to JPEG or, better, video.

The idea behind my RPi 5 was to do all the CV on-robot, but it cut the battery life from 5hrs to 3hrs. So now I have so much compute sitting unused, and such a long runtime, I'm not concerned with only a little image compression while avoiding hardware complexity.

christianrauch:
ROS 2 exposes a couple of [QoS parameters](https://docs.ros.org/en/rolling/Concepts/Intermediate/About-Quality-of-Service-Settings.html) that control the message transport, and then you also have different RMW implementations (DDS, Zenoh, ...) that make a difference in the communication performance. You will likely get better results with optimising the QoS parameters and by switching to `rmw_zenoh_cpp`.

Exposing these parameters makes sense and is useful for different communication needs.

christianrauch:
@eigendude Do you want to pivot this PR towards exposing the QoS settings? Due to the limitation of the `image_transport::CameraPublisher `, I don't plan to support this. But if there is need for tuning the QoS settings:
- https://docs.ros.org/en/jazzy/Concepts/Intermediate/About-Quality-of-Service-Settings.html
- https://docs.ros.org/en/jazzy/Tutorials/Demos/Quality-of-Service.html
then this would be something useful to expose as configurable.

christianrauch:
Would something like https://github.com/christianrauch/camera_ros/pull/155 work for you? It exposes some of the QoS settings for the image publishers. Which QoS settings do you need to adjust?

Issue #147, opened 2025-12-10
Inconclusive behavior if USB camera is disconnected


larsll:
If a USB camera is disconnected and after some time reconnected then the node stops working with some "No such device" messages as expected, but upon reconnecting it only shows that the new cameras are detected, but does not restart publishing messages.

```
[0:45:26.046765537] [4920] ERROR V4L2 v4l2_videodevice.cpp:1781 /dev/video0[29:cap]: Failed to queue buffer 2: No such device
[0:45:26.050432900] [4920] ERROR V4L2 v4l2_videodevice.cpp:1781 /dev/video0[29:cap]: Failed to queue buffer 1: No such device
[0:45:26.051200009] [4920] ERROR V4L2 v4l2_videodevice.cpp:1781 /dev/video0[29:cap]: Failed to queue buffer 3: No such device
[0:45:52.850347453] [4920]  WARN V4L2 v4l2_pixelformat.cpp:346 Unsupported V4L2 pixel format H264
[0:45:52.851563574] [4920]  INFO Camera camera_manager.cpp:220 Adding camera '\_SB_.PCI0.XHC_.RHUB.HS03-3:1.2-29fe:4d53' for pipeline handler uvcvideo
[0:45:52.860260516] [4920]  WARN V4L2 v4l2_pixelformat.cpp:346 Unsupported V4L2 pixel format H264
[0:45:52.861634591] [4920]  INFO Camera camera_manager.cpp:220 Adding camera '\_SB_.PCI0.XHC_.RHUB.HS03-3:1.0-29fe:4d53' for pipeline handler uvcvideo
```

Solution could be to 'redetect' the camera and start publishing again (complicated), or to stop the node when the disconnect is detected as it will no longer function (easy).

larsll:
Closing the issue as #148 was merged. Reconnect functionality to be added separately.

PR #148, opened 2025-12-10
Handle Camera Disconnect

larsll:
This pull request improves the robustness of the `CameraNode` class by adding proper handling for camera disconnections and ensuring resources are cleaned up when the node is shut down. The main changes include introducing an `onDisconnect` handler, connecting and disconnecting the camera's `disconnected` signal.

It is a basic fix to #147 

This pull request adds robust handling for camera disconnections in the `CameraNode` class. The main changes ensure that when the camera disconnects, the node logs a fatal error, publishes a diagnostic message, and safely stops further processing. Additionally, the code now properly manages signal connections and disconnections to prevent resource leaks.

**Camera disconnection handling:**

* Added an `onDisconnect` method to `CameraNode` that logs a fatal error, publishes a diagnostic error message, sets the node as not running, and notifies all condition variables when the camera disconnects.
* Connected the camera's `disconnected` signal to the new `onDisconnect` method in the constructor.
* Disconnected the camera's `disconnected` signal in the destructor to prevent resource leaks.

**Request processing improvements:**

* In `process`, added a check to skip further processing for requests with status `RequestCancelled`.

### DIFF
--- a/src/CameraNode.cpp
+++ b/src/CameraNode.cpp
@@ -121,6 +121,9 @@ private:
   std::atomic_uint8_t jpeg_quality;
 
   void
+  onDisconnect();
+
+  void
   requestComplete(libcamera::Request *const request);
 
   void
@@ -393,6 +396,8 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options)
   if (camera->acquire())
     throw std::runtime_error("failed to acquire camera");
 
+  camera->disconnected.connect(this, &CameraNode::onDisconnect);
+
   std::vector<libcamera::StreamRole> roles {role};
 
   // Add the RAW role if the sensor_size is defined
@@ -605,6 +610,8 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options)
 
 CameraNode::~CameraNode()
 {
+  camera->disconnected.disconnect(this, &CameraNode::onDisconnect);
+
   // stop request callbacks
   camera->requestCompleted.disconnect(this, &CameraNode::requestComplete);
 
@@ -636,6 +643,30 @@ CameraNode::~CameraNode()
   for (const auto &e : buffer_info)
     if (munmap(e.second.data, e.second.size) == -1)
       std::cerr << "munmap failed: " << std::strerror(errno) << std::endl;
+}
+
+void
+CameraNode::onDisconnect()
+{
+  RCLCPP_FATAL_STREAM(get_logger(), "camera '" << camera->id() << "' disconnected!");
+
+  if (pub_diagnostics->get_subscription_count()) {
+    diagnostic_msgs::msg::DiagnosticArray diagnostic_array;
+    diagnostic_array.header.stamp = this->now();
+    diagnostic_array.header.frame_id = frame_id;
+
+    diagnostic_msgs::msg::DiagnosticStatus diagnostic_status;
+    diagnostic_status.hardware_id = camera->id();
+    diagnostic_status.level = diagnostic_msgs::msg::DiagnosticStatus::ERROR;
+    diagnostic_status.message = "camera disconnected";
+
+    diagnostic_array.status.push_back(diagnostic_status);
+    pub_diagnostics->publish(diagnostic_array);
+  }
+
+  running = false;
+  for (auto &[req, condvar] : request_condvars)
+    condvar.notify_all();
 }
 
 void
@@ -780,6 +811,10 @@ CameraNode::process(libcamera::Request *const request)
 
     if (const int ret = camera->queueRequest(request); ret < 0) {
       RCLCPP_WARN_STREAM(get_logger(), "failed to queue request (" << request->toString() << "): " << strerror(-ret));
+      if (ret == -ENODEV || ret == -EACCES) {
+        running = false;
+        return;
+      }
     }
 
     if (!parameter_handler.sync_control_values(request->controls())) {


src/CameraNode.cpp
```diff
@@ -608,6 +620,16 @@ CameraNode::~CameraNode()
...
 }
 
+void
+CameraNode::onDisconnect()
+{
+  RCLCPP_FATAL_STREAM(get_logger(), "camera '" << camera->id() << "' disconnected!");
+  running = false;
+  for (auto &[req, condvar] : request_condvars)
+    condvar.notify_all();
+  rclcpp::shutdown();
```
christianrauch: We should not shut down ROS globally, as this will also shut down all other nodes in the same container. But I am also not sure what the correct behaviour is to unload or quit your own node.

christianrauch (changes requested):
Can you remove the global function calls from the `rclcpp::` namespace and squash your commits?

src/CameraNode.cpp
larsll: Checking back here (forgot all about the PR due to other things) -- reading further rclcpp::shutdown() would indeed stop all nodes within one process, so is probably not good to use in this context.

src/CameraNode.cpp
```diff
@@ -618,7 +640,7 @@ CameraNode::requestComplete(libcamera::Request *const request)
 void
 CameraNode::process(libcamera::Request *const request)
 {
-  while (true) {
+  while (rclcpp::ok() && running) {
```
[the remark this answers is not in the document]
larsll: `rclcpp::ok()` is as far as I read the preferred way to enable a graceful exit from any while(true) style loop in ROS2. In our case `running` has a very similar meaning. Having either of them at the beginning of the loop is prudent to avoid a `process()` loop entering `wait()` after the notifications have been sent.

larsll:
Investigated further -- two scenarios:
- Node launched standalone - `rclcpp::shutdown()` does the desirable thing.
- Node launched using ComposableNode - `rclcpp::shutdown()` tears down all nodes in container. Case dependent if ok or not.

There is apparently no easy way to detect if the node is one or the other. Throwing an exception leads to unclean teardown, which is not desirable, esp. not in a ComposableNode.

Respawn is available to start both standalone and compsable node containers if stopped.

For now the clean way is to have a parameter 'shutdown_on_disconnect' and only run `rclcpp::shutdown()` if true.

christianrauch:
I am still opposed to messing with the container lifecycle like this.

The most sensible way, IMHO, is to notify the user about the disconnect via a log message and via the status change on the "diagnostics" topic. The consumer of this state can then handle the case as they wish. Other nodes and packages, such as `realsense2_camera`, show a warning that no sensor is connected. I would prefer that the `camera_ros` node behaves the same. libcamera has a way to detect online connections of cameras, and in the long-term `camera_ros` should support this online detach and reconnect feature.

larsll:
> The most sensible way, IMHO, is to notify the user about the disconnect via a log message and via the status change on the "diagnostics" topic. The consumer of this state can then handle the case as they wish.

The problem with the approach is that there is no clean way for any node to stop another node without reverting to some ps/kill trickery. Until a the full lifecycle management is implemented it is a choice between two less-than-optimal solutions.

Additionally: The current diagnostics messages are only sent from within the  `process()` loop. If the camera is disconnected then the early return check on `running` will prevent any such message from being sent, hence a suitable location for sending out at least one "status not OK" would have to be found.

christianrauch:
> > The most sensible way, IMHO, is to notify the user about the disconnect via a log message and via the status change on the "diagnostics" topic. The consumer of this state can then handle the case as they wish.
> 
> The problem with the approach is that there is no clean way for any node to stop another node without reverting to some ps/kill trickery. Until a the full lifecycle management is implemented it is a choice between two less-than-optimal solutions.

My point is that we do not have to stop the node. The node could just wait for the camera to reconnect. This is what some other camera nodes do.

> Additionally: The current diagnostics messages are only sent from within the `process()` loop. If the camera is disconnected then the early return check on `running` will prevent any such message from being sent, hence a suitable location for sending out at least one "status not OK" would have to be found.

You can still publish messages outside of the `requestCompleted` callback. So when the disconnect event is triggered, you could publish the diagnostics message.

larsll:
This is now the minimum version. There are an error message displayed and a diagnostics message sent. 

Having to monitor `/diagostics` to find a potential error is not really that great, as there might be a lot of messages flowing -- including one message per camera frame -- causing significant load on the node that has to process it.

src/CameraNode.cpp
```diff
@@ -605,6 +610,8 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options)
 
 CameraNode::~CameraNode()
 {
+  camera->disconnected.disconnect(this);
```
christianrauch: Don't you have to provide the `&CameraNode::onDisconnect` callback here? Have you checked that this indeed disconnects the callback without providing the function specifically?

src/CameraNode.cpp
```diff
@@ -766,6 +797,9 @@ CameraNode::process(libcamera::Request *const request)
 
     pub_diagnostics->publish(diagnostic_array);
 
+    if (request->status() == libcamera::Request::RequestCancelled)
+      continue;
```
christianrauch: This way, you skip over the `request->reuse()` and the requests won't be re-used again. Doesn't this mean the node stops indefinitely to process requests, even if we reconnect the camera? Have you checked that the while loop continues when the camera is re-attached?

Also, the `Request::RequestCancelled` condition is already handled in the if/else and it is meant to not stop the processing, as the camera can recover from this. I.e. a single cancelled frame should never stop the processing of the remaining frames.

christianrauch:
> Having to monitor `/diagostics` to find a potential error is not really that great, as there might be a lot of messages flowing -- including one message per camera frame -- causing significant load on the node that has to process it.

Isn't this standard in ROS? From the diagnostics level you can easily see when something is not right. Compared to the bandwidth for the images, the diagnostics only contain a few words of text and should not really impact the bandwidth or node usage. There are sensors with way higher update rates.

If you find alternative solutions for explicitly sending "sensor connection" states, i.e. from other common ROS sensor nodes, we can also see about integrating those.

src/CameraNode.cpp
christianrauch: To give you a concrete example, requests are cancelled when they are configured with auto exposure and a fixed exposure time. This can be rectified easily by adjusting the parameters online. Stopping the processing entirely would not allow such a recovery.

larsll: Ok. Good catch. There needs to be a differentiation between a disconnected camera cancellation and any other cancellation. Trying to re-use a request when the camera is disconnected leads to a core dump. (Which is why this protection was added.)

When disconnecting I got:

```
[0:09:39.774780692] [1835] ERROR V4L2 v4l2_videodevice.cpp:1789 /dev/video0[29:cap]: Failed to queue buffer 0: No such device
[0:09:39.778733616] [1835] ERROR V4L2 v4l2_videodevice.cpp:1789 /dev/video0[29:cap]: Failed to queue buffer 2: No such device
[0:09:39.779293191] [1835] ERROR V4L2 v4l2_videodevice.cpp:1789 /dev/video0[29:cap]: Failed to queue buffer 1: No such device
[0:09:39.780157472] [1835] ERROR V4L2 v4l2_videodevice.cpp:1789 /dev/video0[29:cap]: Failed to queue buffer 3: No such device
[WARN] [1784400736.964844687] [camera]: sensor timestamp not available, falling back to node time as reference
[ERROR] [1784400736.966788802] [camera]: request 'Request(655:X:0/1:0)' cancelled
[ERROR] [1784400736.966934667] [camera]: request 'Request(653:X:0/1:0)' cancelled
[ERROR] [1784400736.966998903] [camera]: request 'Request(652:X:0/1:0)' cancelled
[ERROR] [1784400736.967512318] [camera]: request 'Request(654:X:0/1:0)' cancelled
[0:09:39.790155734] [1835] ERROR V4L2 v4l2_videodevice.cpp:1789 /dev/video0[29:cap]: Failed to queue buffer 3: No such device
[0:09:39.790590730] [1835] ERROR V4L2 v4l2_videodevice.cpp:1789 /dev/video0[29:cap]: Failed to queue buffer 2: No such device
[0:09:39.791305581] [1835] ERROR V4L2 v4l2_videodevice.cpp:1789 /dev/video0[29:cap]: Failed to queue buffer 0: No such device
[0:09:39.791950060] [1835] ERROR V4L2 v4l2_videodevice.cpp:1789 /dev/video0[29:cap]: Failed to queue buffer 1: No such device
[ERROR] [1784400736.976242595] [camera]: request 'Request(656:X:0/1:0)' cancelled
[ERROR] [1784400736.976982481] [camera]: request 'Request(657:X:0/1:0)' cancelled
[0:09:39.796196627] [1835] ERROR V4L2 v4l2_videodevice.cpp:1789 /dev/video0[29:cap]: Failed to queue buffer 3: No such device
[ERROR] [1784400736.977059107] [camera]: request 'Request(658:X:0/1:0)' cancelled
[0:09:39.801256924] [1835] ERROR V4L2 v4l2_videodevice.cpp:1789 /dev/video0[29:cap]: Failed to queue buffer 2: No such device
[0:09:39.801422096] [1835] ERROR V4L2 v4l2_videodevice.cpp:1789 /dev/video0[29:cap]: Failed to queue buffer 0: No such device
[ERROR] [1784400736.982308679] [camera]: request 'Request(659:X:0/1:0)' cancelled
[ERROR] [1784400736.991892509] [camera]: request 'Request(660:X:0/1:0)' cancelled
[ERROR] [1784400736.993701578] [camera]: request 'Request(661:X:0/1:0)' cancelled
[ERROR] [1784400736.994377248] [camera]: request 'Request(662:X:0/1:0)' cancelled
[FATAL] [1784400736.996337398] [camera]: camera '\_SB_.PCI0.XHC_.RHUB.HS03-3:1.0-29fe:4d53' disconnected!
[0:09:39.815079449] [1835] ERROR V4L2 v4l2_videodevice.cpp:1789 /dev/video0[29:cap]: Failed to queue buffer 1: No such device
[0:09:39.817145055] [1835] ERROR Camera camera.cpp:723 Camera in Configured state trying requestComplete() requiring state between Stopping and Running
[0:09:39.818479712] [1835] FATAL Camera camera.cpp:1487 Trying to complete a request when stopped
Backtrace:
/opt/ros/jazzy/lib/libcamera.so.0.7(_ZN9libcamera6Camera15requestCompleteEPNS_7RequestE+0x29d) [0x7edf546fdcdd]
/opt/ros/jazzy/lib/libcamera.so.0.7(_ZN9libcamera15PipelineHandler15completeRequestEPNS_7RequestE+0x66) [0x7edf54731776]
/opt/ros/jazzy/lib/libcamera.so.0.7(_ZN9libcamera15PipelineHandler15doQueueRequestsEPNS_6CameraE+0x43) [0x7edf54731a73]
/opt/ros/jazzy/lib/libcamera-base.so.0.7(_ZN9libcamera15BoundMethodBase12activatePackESt10shared_ptrINS_19BoundMethodPackBaseEEb+0x50) [0x7edf55dbb470]
/opt/ros/jazzy/lib/libcamera.so.0.7(+0xbeb8b) [0x7edf54731b8b]
/opt/ros/jazzy/lib/libcamera.so.0.7(_ZN9libcamera7Request7Private20emitPrepareCompletedEv+0x4c) [0x7edf5470a2dc]
/opt/ros/jazzy/lib/libcamera.so.0.7(_ZN9libcamera15PipelineHandler12queueRequestEPNS_7RequestE+0x48) [0x7edf54731618]
/opt/ros/jazzy/lib/libcamera-base.so.0.7(_ZN9libcamera6Object7messageEPNS_7MessageE+0x55) [0x7edf55dba9c5]
/opt/ros/jazzy/lib/libcamera-base.so.0.7(_ZN9libcamera6Thread16dispatchMessagesENS_7Message4TypeEPNS_6ObjectE+0x195) [0x7edf55dbc295]
/opt/ros/jazzy/lib/libcamera-base.so.0.7(_ZN9libcamera19EventDispatcherPoll13processEventsEv+0x3b) [0x7edf55dc360b]
/opt/ros/jazzy/lib/libcamera-base.so.0.7(_ZN9libcamera6Thread4execEv+0x69) [0x7edf55dc2319]
/opt/ros/jazzy/lib/libcamera.so.0.7(_ZN9libcamera13CameraManager7Private3runEv+0x84) [0x7edf54701684]
/lib/x86_64-linux-gnu/libstdc++.so.6(+0xecdb4) [0x7edf57eecdb4]
/lib/x86_64-linux-gnu/libc.so.6(+0x9caa4) [0x7edf57a9caa4]
/lib/x86_64-linux-gnu/libc.so.6(+0x129c6c) [0x7edf57b29c6c]
[ros2run]: Aborted
```

larsll:
> > Having to monitor `/diagostics` to find a potential error is not really that great, as there might be a lot of messages flowing -- including one message per camera frame -- causing significant load on the node that has to process it.
> 
> Isn't this standard in ROS? From the diagnostics level you can easily see when something is not right. Compared to the bandwidth for the images, the diagnostics only contain a few words of text and should not really impact the bandwidth or node usage. There are sensors with way higher update rates.
> 
> If you find alternative solutions for explicitly sending "sensor connection" states, i.e. from other common ROS sensor nodes, we can also see about integrating those.

I think it is the standard way for human diagnostics; but not sure if it is the standard way for node-to-node communication, as there is no easy way to filter out messages, so depending on how many other nodes are also publishing joint `/diagnostics` this is the more or less work the node needs to do. Esp. in Python this can be 'heavy'.

src/CameraNode.cpp
larsll: https://docs.libcamera.org/master/public-api/classlibcamera_1_1Signal.html says that all slots will then be disconnected, Still, will add `&CameraNode::onDisconnect` to match the pattern for request callbacks two lines down.

larsll:
This should now work as intended, an `FATAL` message is displayed and the node goes dormant.
 #173 will look at handling the reconnection.

src/CameraNode.cpp
christianrauch: I was blindly assuming that the callbacks are not called anymore when a camera is disconnected.

Is there an official example of how to handle the `disconnected` signal?

larsll: I did not find any -- on https://docs.libcamera.org/master/public-api/classlibcamera_1_1Camera.html#ab4c410debecced9a862f856fa3005282 there is only a comment 

> As soon as this signal is emitted the camera instance will refuse all new application API calls by returning errors immediately.

I suspect there is the potential of a race condition if a process-thread is working on a frame during the disconnect. I could not reproduce a core dump after I added the protection `if (ret == -ENODEV || ret == -EACCES)`, which then sets `running = false`.

christianrauch (approved):

PR #173, opened 2026-07-18
Camera Reconnect

larsll:
This PR adds reconnect handling so the node can recover when libcamera reports that the active camera disappeared, without requiring a full node restart.

## Design

The key constraint is that libcamera fires both `disconnected` and `cameraAdded` signals on its internal worker thread. Doing any heavy work there (thread joins, `camera->stop()`, `camera->release()`, `camera->acquire()`, `configure()`) causes crashes. Both signal handlers therefore only set `std::atomic` flags and log/publish diagnostics; all actual teardown and restart is deferred to a 200 ms `rclcpp::wall_timer` (`onReconnectTimer`) that runs on the node's executor thread.

## Reconnect flow

**On disconnect** (`onDisconnect`, libcamera signal thread):
- Sets `running = false` and `camera_disconnected = true`.
- Logs the disconnect using the cached `camera_hardware_id` (no `camera` pointer dereference).
- Publishes an ERROR diagnostic.

**On camera added** (`onCameraAdded`, libcamera signal thread):
- Returns immediately if `camera_hardware_id` is empty (no camera was ever fully acquired).
- If the added camera's libcamera ID does not match `camera_hardware_id`, logs a WARN and ignores it — only the same physical camera at the same port/bus address triggers a reconnect.
- If the ID matches, sets `reconnect_requested = true`.

**On timer tick** (`onReconnectTimer`, executor thread):
- If `camera_disconnected` is set: calls `stopCamera(disconnected=true)`.
- If no camera is held and `reconnect_requested` is set: calls `startCamera()` and publishes an OK diagnostic on success, or logs the error and calls `stopCamera()` on failure.

## Teardown (`stopCamera`)

`stopCamera` accepts a `disconnected` flag. When `true` (hardware already gone):
- Skips `camera->stop()`, `allocator->free()`, and `camera->release()` — calling into libcamera against a removed device dereferences torn-down state and crashes.
- Drops the node's shared-pointer references only; libcamera emits a benign _"Removing camera while still in use"_ message during this, which is expected and safe.
- Memory-mapped buffers (`mmap`) are always unmapped regardless of the flag.

## Restart (`startCamera`)

On reconnect, `startCamera` uses `camera_manager.get(camera_hardware_id)` to re-acquire the same physical camera by its stored libcamera ID. `camera_hardware_id` is written once on the first successful connection and kept across disconnects as the reconnect anchor. `parameter_handler.declare()` is only called on first connection and is not repeated on reconnect, preserving existing declared ROS parameters.

## Other changes

- `request_state_mutex` guards the request mutex/condvar maps so `requestComplete()` cannot race with `stopCamera()` clearing those maps mid-teardown.
- Queue failures with `-ENODEV` or `-EACCES` in `process()` set `running = false` and exit the processing loop instead of continuing against a dead camera.
- Per-frame diagnostics populate `hardware_id` from the cached `camera_hardware_id`, avoiding any dereference of `camera` from processing threads during or after a disconnect.

## Before merging

- [x] Test disconnect/reconnect repeatedly on a USB/UVC camera.
- [ ] Test normal functionality on a Raspberry Pi/libcamera native camera.
- [x] Test normal node shutdown with a connected camera.
- [x] Test shutdown after the camera has already disconnected.
- [x] Test reconnect with the default camera selection.
- [x] Test reconnect with a specific `camera` parameter configured, including ignoring non-matching added cameras.
- [ ] Test on ROS 2 Humble.
- [x] Test on at least one newer ROS 2 release.
- [x] Confirm camera controls still apply after reconnect for cameras that expose them.
- [x] Confirm cameras that do not expose optional controls, such as `FrameDurationLimits` on some USB cameras, behave gracefully.
- [x] Check diagnostics: ERROR on disconnect and OK on successful reconnect.
- [x] Confirm CI/build passes across the supported libcamera/ROS matrix.

### DIFF
--- a/src/CameraNode.cpp
+++ b/src/CameraNode.cpp
@@ -15,7 +15,9 @@
 #include <cv_bridge/cv_bridge.h>
 #endif
 #include <atomic>
+#include <chrono>
 #include <diagnostic_msgs/msg/diagnostic_array.hpp>
+#include <functional>
 #include <iostream>
 #include <libcamera/base/shared_fd.h>
 #include <libcamera/base/signal.h>
@@ -32,6 +34,7 @@
 #include <libcamera/request.h>
 #include <libcamera/stream.h>
 #include <memory>
+#include <mutex>
 #include <opencv2/core/mat.hpp>
 #include <opencv2/imgcodecs.hpp>
 #include <optional>
@@ -87,7 +90,10 @@ private:
   std::vector<std::thread> request_threads;
   std::unordered_map<const libcamera::Request *, std::mutex> request_mutexes;
   std::unordered_map<const libcamera::Request *, std::condition_variable> request_condvars;
+  std::mutex camera_state_mutex;
   std::atomic<bool> running;
+  std::atomic<bool> camera_disconnected;
+  std::atomic<bool> reconnect_requested;
 
   struct buffer_info_t
   {
@@ -97,6 +103,7 @@ private:
   std::unordered_map<const libcamera::FrameBuffer *, buffer_info_t> buffer_info;
 
   bool use_node_time;
+  std::string camera_hardware_id;
 
   static const rclcpp::PublisherOptionsWithAllocator<std::allocator<void>> pubopts;
 
@@ -120,8 +127,23 @@ private:
   // compression quality parameter
   std::atomic_uint8_t jpeg_quality;
 
+  // defers camera teardown/restart off libcamera's signal-callback thread
+  rclcpp::TimerBase::SharedPtr reconnect_timer;
+
   void
   onDisconnect();
+
+  void
+  onCameraAdded(std::shared_ptr<libcamera::Camera> added_camera);
+
+  void
+  startCamera();
+
+  void
+  stopCamera(const bool disconnected = false);
+
+  void
+  onReconnectTimer();
 
   void
   requestComplete(libcamera::Request *const request);
@@ -236,6 +258,10 @@ compressImageMsg(const sensor_msgs::msg::Image &source,
 
 CameraNode::CameraNode(const rclcpp::NodeOptions &options)
     : Node("camera", options),
+      stream(nullptr),
+      running(false),
+      camera_disconnected(false),
+      reconnect_requested(false),
 #if CIM_HAS_NODE_INTERFACE
       cim(
         this->get_node_base_interface(),
@@ -269,22 +295,20 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options)
   param_descr_role.description = "stream role";
   param_descr_role.additional_constraints = "one of {raw, still, video, viewfinder}";
   param_descr_role.read_only = true;
-  const libcamera::StreamRole role =
-    get_role(declare_parameter<std::string>("role", "viewfinder", param_descr_role));
+  declare_parameter<std::string>("role", "viewfinder", param_descr_role);
 
   // image dimensions
   rcl_interfaces::msg::ParameterDescriptor param_descr_ro;
   param_descr_ro.read_only = true;
-  const uint32_t w = declare_parameter<int64_t>("width", {}, param_descr_ro);
-  const uint32_t h = declare_parameter<int64_t>("height", {}, param_descr_ro);
-  const libcamera::Size size {w, h};
+  declare_parameter<int64_t>("width", {}, param_descr_ro);
+  declare_parameter<int64_t>("height", {}, param_descr_ro);
 
   // Raw format dimensions
   rcl_interfaces::msg::ParameterDescriptor param_descr_sensor_mode;
   param_descr_sensor_mode.description = "raw mode of the sensor";
   param_descr_sensor_mode.additional_constraints = "string in format [width]:[height]";
   param_descr_sensor_mode.read_only = true;
-  const libcamera::Size sensor_size = get_sensor_format(declare_parameter<std::string>("sensor_mode", {}, param_descr_sensor_mode));
+  declare_parameter<std::string>("sensor_mode", {}, param_descr_sensor_mode);
 
   // camera frame_id
   frame_id = declare_parameter<std::string>("frame_id", "camera", param_descr_ro);
@@ -298,10 +322,9 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options)
   param_descr_orientation.integer_range.push_back(orientation_range);
   param_descr_orientation.read_only = true;
   constexpr int orientation_angle_default = 0;
-  const int angle = declare_parameter<int>("orientation", orientation_angle_default, param_descr_orientation);
-#if LIBCAMERA_VER_GE(0, 2, 0)
-  const libcamera::Orientation orientation = libcamera::orientationFromRotation(angle);
-#else
+  declare_parameter<int>("orientation", orientation_angle_default, param_descr_orientation);
+#if !LIBCAMERA_VER_GE(0, 2, 0)
+  const int angle = get_parameter("orientation").as_int();
   if (angle != orientation_angle_default) {
     RCLCPP_WARN_STREAM(get_logger(), "parameter 'orientation' not supported on libcamera " << LIBCAMERA_VERSION_MAJOR << "." << LIBCAMERA_VERSION_MINOR);
   }
@@ -311,10 +334,10 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options)
   rcl_interfaces::msg::ParameterDescriptor param_descr_camera_info_url;
   param_descr_camera_info_url.description = "camera calibration info file url";
   param_descr_camera_info_url.read_only = true;
+  declare_parameter<std::string>("camera_info_url", {}, param_descr_camera_info_url);
 
   // camera ID
-  const rclcpp::ParameterValue &camera_id =
-    declare_parameter("camera", rclcpp::ParameterValue {}, param_descr_ro.set__dynamic_typing(true));
+  declare_parameter("camera", rclcpp::ParameterValue {}, param_descr_ro.set__dynamic_typing(true));
 
   // we cannot control the compression rate of the libcamera MJPEG stream
   // ignore "jpeg_quality" parameter for MJPEG streams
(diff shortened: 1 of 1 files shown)


larsll:
@christianrauch This is now working for basic uvcvideo camera. Still some different testing needed to ensure that it is mergeable as the change is quite big.

I would request your review on structure of things before going down the testing track.

src/CameraNode.cpp
```diff
@@ -262,29 +289,29 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options)
   rcl_interfaces::msg::ParameterDescriptor param_descr_format;
   param_descr_format.description = "pixel format of streaming buffer";
   param_descr_format.read_only = true;
-  const std::string &format = declare_parameter<std::string>("format", {}, param_descr_format);
+  format = declare_parameter<std::string>("format", {}, param_descr_format);
```
christianrauch: How are these changes, and the ones below, related to the camera reconnect detection? This shouldn't be required. Also, the parameter is `read_only`, so the variable should also be `const`.

src/CameraNode.cpp
```diff
@@ -608,78 +701,108 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options)
...
     if (munmap(e.second.data, e.second.size) == -1)
       std::cerr << "munmap failed: " << std::strerror(errno) << std::endl;
+  buffer_info.clear();
+  stream = nullptr;
+  camera_hardware_id.clear();
 }
 
 void
-CameraNode::onDisconnect()
+CameraNode::publishDiagnosticStatus(const uint8_t level, const std::string &message)
```
christianrauch: This looks like a refactoring commit that is also unrelated to the reconnect functionality. I would prefer if the "refactoring" changes that do not implement new functionality are maintained in separate commits or PRs.

src/CameraNode.cpp
```diff
@@ -608,78 +701,108 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options)
...
+
+  std::unique_lock lk(mutex_it->second);
+  condvar_it->second.notify_one();
 }
 
 void
 CameraNode::process(libcamera::Request *const request)
 {
-  while (true) {
+  while (rclcpp::ok()) {
```
christianrauch: Shouldn't this work without the `rclcpp::ok()`?

src/ParameterHandler.cpp
```diff
@@ -71,6 +71,9 @@ ParameterHandler::declare(const libcamera::ControlInfoMap &controls)
   // to unset them (set their type to 'rclcpp::ParameterType::PARAMETER_NOT_SET').
   // Unsetting a statically typed parameter causes "cannot undeclare a statically typed parameter".
 
+  camera_controls.clear();
+  parameter_descriptors.clear();
```
christianrauch: Why is this required? Should the controls be the same after the disconnected camera reconnects?

christianrauch (changes requested):

src/CameraNode.cpp
larsll: The variables now need to be stored so that they can be re-used when a camera is reconnected. Creating them `const` is doable only with a helper function in the constructor initializer list which will require a quite substantial re-do of the constructor...

src/ParameterHandler.cpp
larsll: If it is the same camera, then I assume yes. If it is a different one then no. Prudent to refresh all of it.

src/CameraNode.cpp
larsll: It snuck over from the other branch. I will remove it.

src/CameraNode.cpp
larsll: The publishDiagnosticStatus was added as the message is being sent both on connect and on disconnect.

src/ParameterHandler.cpp
christianrauch: I haven't thought about connecting a different camera.

In this case, we should probably re-instantiate the `ParameterHandler` instance to make sure that everything else (`control_values`, the mutexes, ...) is also reset.

larsll: It is obviously an option to block a different device from connecting, as there is the risk that the provided parameters (dimensions, pixel format, sensor format etc.) are not compatible with the new camera.

src/CameraNode.cpp
christianrauch: So when you connect a different camera, you need to reconfigure it with these parameters. I am not sure if this is something we can support currently. If a user configures the node with a format that is incompatible with the camera, it will throw an exception, and the node will not start. The user can then adjust the configuration to make this work for a camera. How will this work, wenn the user connects a camera online that does not support the initially valid configuration?

I think we should only support reconnecting the initial camera and not support the more complex "any camera" unless someone needs this more complex "connect any camera at any time" feature.

larsll: Ok. Let's do that. I will alter the code accordingly.

larsll:
@christianrauch This is now implemented with the limitation that only the same camera can be reinserted. This also comes with the limitation that only 'same camera, same port' works in the USB scenario.

The PR is a bit bigger than desired, but the timer seems to be required to avoid intermittent segfaults (10%-20% of disconnects) inside of libcamera when cleaning things up.

christianrauch:
The total diff is quite extensive now, and it is hard to understand what the actual functional changes are. For example, the diff shows that `CameraNode::~CameraNode()` is replaced by `void CameraNode::stopCamera()`, and `CameraNode::onDisconnect()` replaced by `CameraNode::publishDiagnosticStatus()`.

Is it possible to split this into more sensible commits to make this easier to read? E.g. split the refactoring from the actual functional changes? Also, can you merge the fixup commits so this can be reviewed in smaller chunks, commit-by-commit?

larsll:
> The total diff is quite extensive now, and it is hard to understand what the actual functional changes are. For example, the diff shows that `CameraNode::~CameraNode()` is replaced by `void CameraNode::stopCamera()`, and `CameraNode::onDisconnect()` replaced by `CameraNode::publishDiagnosticStatus()`.
> 
> Is it possible to split this into more sensible commits to make this easier to read? E.g. split the refactoring from the actual functional changes? Also, can you merge the fixup commits so this can be reviewed in smaller chunks, commit-by-commit?

I have worked on simplifying the diff now (changing the order of some methods helps a lot)...

The work summarized:
- Constructor gets split almost in the middle; the camera connection part goes into `void CameraNode::startCamera()`.
- `CameraNode::onDisconnect()` and  `CameraNode::~CameraNode()`  functionality goes into `void CameraNode::stopCamera()`.
- I removed `CameraNode::publishDiagnosticStatus()` for now -- the two calls are inline. We can refactor it out in a next step.

src/CameraNode.cpp
```diff
@@ -354,40 +377,72 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options)
...
+  const std::string format = get_parameter("format").as_string();
+  const libcamera::StreamRole role = get_role(get_parameter("role").as_string());
+  const libcamera::Size size {
+    static_cast<uint32_t>(get_parameter("width").as_int()),
+    static_cast<uint32_t>(get_parameter("height").as_int())};
+  const libcamera::Size sensor_size = get_sensor_format(get_parameter("sensor_mode").as_string());
+#if LIBCAMERA_VER_GE(0, 2, 0)
+  const libcamera::Orientation orientation = libcamera::orientationFromRotation(static_cast<int>(get_parameter("orientation").as_int()));
+#endif
+  const rclcpp::ParameterValue camera_id = get_parameter("camera").get_parameter_value();
```
christianrauch, line 406: Since these parameters are read-only, I suggest that they are read and stored when the node is constructed and that they are then reused when `startCamera` is called. This removes the repetition of the parameter name and also avoids the expensive API calls by caching the read-only values. There is no reason we have to call `get_parameter` repeatedly, when the value is not going to change.

Also note that all this happens with the locked mutex, which should be avoided. There is no need to lock the mutex for these parameters, as their value is never going to change.

src/CameraNode.cpp
```diff
@@ -269,22 +295,20 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options)
...
+  declare_parameter<int64_t>("width", {}, param_descr_ro);
+  declare_parameter<int64_t>("height", {}, param_descr_ro);
 
   // Raw format dimensions
   rcl_interfaces::msg::ParameterDescriptor param_descr_sensor_mode;
   param_descr_sensor_mode.description = "raw mode of the sensor";
   param_descr_sensor_mode.additional_constraints = "string in format [width]:[height]";
   param_descr_sensor_mode.read_only = true;
-  const libcamera::Size sensor_size = get_sensor_format(declare_parameter<std::string>("sensor_mode", {}, param_descr_sensor_mode));
+  declare_parameter<std::string>("sensor_mode", {}, param_descr_sensor_mode);
```
christianrauch, line 311: We should keep reading the parameter value here and store it in a variable for later use. See comment below.

src/CameraNode.cpp
```diff
@@ -354,40 +377,72 @@ CameraNode::CameraNode(const rclcpp::NodeOptions &options)
...
-  case rclcpp::ParameterType::PARAMETER_NOT_SET:
+  if (!camera_hardware_id.empty()) {
+    // reconnect: find the same physical camera by its hardware ID
+    camera = camera_manager.get(camera_hardware_id);
+    if (!camera) {
+      RCLCPP_INFO_STREAM(get_logger(), camera_manager);
+      throw std::runtime_error("camera '" + camera_hardware_id + "' not found");
+    }
+  }
+  else if (camera_id.get_type() == rclcpp::ParameterType::PARAMETER_NOT_SET) {
```
christianrauch, line 417: Why not keep the switch / case here? This refactoring seems unrelated to the reconnect change.

christianrauch (changes requested):
The diff mostly looks good. I haven't tested the functionality yet. There are some unnecessary changes related to parameter reading and the switch for the parameter type. Please only add changes that are absolutely required for adding the functionality. Do not add style or refactoring changes, such as replacing the switch/case with if/else.

Also, the commit message does not make any sense at all. Is this AI generated? Please have a look at tutorials how to format git commit messages.

src/CameraNode.cpp
larsll: It is a balancing act - the repeated calls to get_parameter (on the rare reconnect), or the creation of a set of variables that are permanently kept in memory.

What is the preferred approach:
- Non-const variables assigned in the constructor
- const variables + helper functions to set them before the real constructor `{}` runs
- A non-struct containing all the variables
- ...

src/CameraNode.cpp
larsll: Here chosen to keep the actual diff small -- in the new if/if-else/else we are checking different variables, not only the `camera_id.get_type()`; and only one of the blocks should run.

Other option is to wrap everything in an if/else -- but in my view it makes the code harder to read.

src/CameraNode.cpp
christianrauch: The parameters are stored in the node anyway, and you can actually get a reference to them if they are too large. In this case, we are just storing a few small variables.

Using non-const class members is the easiest readable variant. For simplicity, I would also create a struct that stores all the parameters that you need to re-initialise the camera when it is re-connected.

src/CameraNode.cpp
christianrauch: But all the `camera_id.get_type()` checks can stay as switch/case, no? You can check for `camera_hardware_id.empty()` separately.

